Issue #23, opened 2026-02-13
Paginate products in the library


binary-koan:
We currently use an endpoint that fetches all library products at once, which is bad for app performance and can fail for large libraries.

The Gumroad endpoint actually supports `per_page` and `page` params for pagination, so we should switch to using those.

A complication is that we use the response to build the creator filter, so we'll need to either create a new endpoint for that data or add it as metadata to the existing endpoint. Either way, we should ideally use the same logic the web app does for the creator filter.

binary-koan:
I think we'll move forward with #69, the approach is good so I think it'll be easier for me to tidy that one up than compare with other PRs sorry

PR #31, opened 2026-02-18
Add pagination and fix creator filtering

Devwill2:
ref #23

# Description

Optimizes the library view by introducing server-side pagination and a dedicated creator filter endpoint to fix performance bottlenecks and filtering bugs on mobile.

## Problem

- **Performance Issues**: The app was fetching the entire library history at once, causing hangs and crashes for users with many purchases.

## Solution

**Backend Changes:**
- Created a lightweight `GET /api/mobile/purchases/creators` endpoint to fetch the full creator list efficiently using SQL aggregations.

**Frontend Changes:**
- Replaced monolithic fetch with `useInfiniteQuery` for efficient, paginated scrolling.
- Converted creator filter state from `Set` to `Array` to fix a React Query bug where filters wouldn't trigger a refetch.
- Centralized API URLs into a new `lib/api-endpoints.ts` file for better maintainability.

# Test Results

All tests passing:
Screenshot 2026-02-19 at 2 07 56 AM

## Video

https://github.com/user-attachments/assets/13bd38c2-c78b-4e27-96aa-1124fcfbe2ed

# AI Disclosure

AI was used to assist with:
- Model: Claude Sonnet 4.5
- IDE : Cursor
- Code refactoring and simplification suggestions


binary-koan (commented):
Thanks for this! Just realised (from looking at how the web library works) that we should be updating the creator counts based on any other filters (search, archived or not). We'll also need to support those parameters in the purchases API, so it may actually make more sense to return the counts in that same API call rather than adding an API

Issue #25, opened 2026-02-13
Support biometric login


binary-koan:
After logging in once, the user should be able to use face/fingerprint/etc to log in again if their session expires. We can use Expo's biometric auth library for this.

PR #32, opened 2026-02-19
feat: Support biometric login (#25)

pegasus1134:
Issue: #25
                                                                                                                                                                                                   
  ## Problem                                                                                                                                                                                     
                                                                                                                                                                                                   
  When a session expires, the only way back in is the full browser OAuth flow. On mobile this is pretty painful: tap sign in, wait for the browser, enter credentials, authorize. Wanted a faster path.

  ## Solution

  After the first OAuth login, the app prompts to enable Face ID / fingerprint. On subsequent session expiry, it uses biometric verification to refresh the session with the stored refresh token... no browser needed.

  The main architectural decision was separating session expiry from explicit logout. Session expiry (401 responses) clears only the access token but keeps the refresh token around for biometric re-auth. Explicit logout from settings wipes everything including biometric preferences. Without this distinction, there's nothing to authenticate against when the user comes back.

  `refreshTokenFn` returns the new access token directly instead of relying on React state - the `accessToken` in the closure is still `null` when `loginWithBiometrics` calls it (React batches state updates), so reading the return value avoids a bug where creators would lose their Dashboard and Analytics tabs after biometric re-login.

  The login screen auto-triggers biometric when available. If the refresh token is expired and fails, `hasRefreshToken` gets cleared immediately in both the null-return and catch paths, which sets `canUseBiometric` to false and breaks the auto-trigger cycle. The user falls through to the normal OAuth button cleanly.

  The biometric utility module (`lib/biometric.ts`) wraps `expo-local-authentication` and `expo-secure-store` - hardware checks, preference persistence, and the authentication prompt with `disableDeviceFallback: true`.

  No Maestro e2e tests included since biometric requires hardware enrollment that Maestro can't trigger on simulators. A settings toggle for enabling/disabling biometric could be a follow-up if desired. The feature is atomic and can't be split without losing testability.

  ## Screenshots                              
                                                                                                                                                                                                   
  ### Light mode
                                                                                                                                                                                                   
  | Before | After |                                                                                                                                                                             
  |--------|-------|                                                                                                                                                                               
  | <img width="300" alt="before-light" src="https://github.com/user-attachments/assets/5ab38358-6f38-4580-b49e-3b640d4255c6" /> | <img width="300" alt="after-light" src="https://github.com/user-attachments/assets/f55a0095-6fa9-4634-b9c4-1615d4395c8a" /> |

  ### Dark mode

  | Before | After |
  |--------|-------|
  | <img width="300" alt="before-dark" src="https://github.com/user-attachments/assets/d2c8719b-3966-4631-bef2-ae40997065bb" /> | <img width="300" alt="after-dark" src="https://github.com/user-attachments/assets/39863460-7434-43a6-b279-252d68787f5c" /> |

  ## Demo Video

 Shows: OAuth login → biometric enrollment prompt → session expiry → auto Face ID on return → successful biometric re-login. Session expiry was simulated to demonstrate the flow. In production this triggers on 401 or token expiry.


https://github.com/user-attachments/assets/8e0152be-a43b-4465-a01c-483f4e416f1a



  ## Test results

25 new tests across 3 suites covering the biometric utility (hardware checks, preference storage, type detection), auth context integration (happy path, failed refresh, cancelled biometric, loop prevention, silent refresh on 401), and login screen rendering (button visibility, biometric-specific labels, auto-trigger behavior). All tests fail when the application code is reverted.

<img width="926" height="1037" alt="Screenshot 2026-02-19 at 11 01 51 PM" src="https://github.com/user-attachments/assets/494d1fdf-c34e-4cdb-9fcb-4a4cc5cd7fd1" />
<img width="1114" height="660" alt="Screenshot 2026-02-19 at 11 02 03 PM" src="https://github.com/user-attachments/assets/9c5e3b50-5c52-4052-9f3c-0d4ceeac44b3" />


  ## Checklist

  - [x] Screenshots attached

  ## AI disclosure

  **Model:** Claude Opus 4.6 via Claude Code CLI

  **Used for:**
  - Analyzing the existing auth flow and session management
  - Running test suites, type checking, linting
  - Identifying edge cases

  Implementation direction and architectural decisions were my own. Session expiry vs logout separation, biometric enrollment UX, refresh token preservation strategy, and UI layout. All code reviewed line-by-line before committing.

### DIFF
--- a/lib/biometric.ts
+++ b/lib/biometric.ts
@@ -1,5 +1,6 @@
 import * as LocalAuthentication from "expo-local-authentication";
 import * as SecureStore from "expo-secure-store";
+import { Platform } from "react-native";
 
 const BIOMETRIC_ENABLED_KEY = "gumroad_biometric_enabled";
 
@@ -20,6 +21,23 @@ export const setBiometricEnabled = async (enabled: boolean): Promise<void> => {
   } else {
     await SecureStore.deleteItemAsync(BIOMETRIC_ENABLED_KEY);
   }
+};
+
+export const getBiometricLabel = async (): Promise<{ label: string; icon: "scan" | "fingerprint" }> => {
+  const types = await LocalAuthentication.supportedAuthenticationTypesAsync();
+  if (types.includes(LocalAuthentication.AuthenticationType.FACIAL_RECOGNITION)) {
+    return {
+      label: Platform.OS === "ios" ? "Face ID" : "face recognition",
+      icon: "scan",
+    };
+  }
+  if (types.includes(LocalAuthentication.AuthenticationType.FINGERPRINT)) {
+    return {
+      label: Platform.OS === "ios" ? "Touch ID" : "fingerprint",
+      icon: "fingerprint",
+    };
+  }
+  return { label: "biometrics", icon: "fingerprint" };
 };
 
 export const authenticate = async (): Promise<boolean> => {
--- a/app/login.tsx
+++ b/app/login.tsx
@@ -13,7 +13,8 @@ import { StyledImage } from "../components/styled";
 import { useAuth } from "../lib/auth-context";
 
 export default function LoginScreen() {
-  const { isAuthenticated, isLoading, login, canUseBiometric, loginWithBiometrics } = useAuth();
+  const { isAuthenticated, isLoading, login, isBiometricSetUp, biometricLabel, biometricIcon, loginWithBiometrics } =
+    useAuth();
   const { theme } = useUniwind();
 
   useEffect(() => {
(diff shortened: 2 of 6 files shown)


lib/biometric.ts
```diff
@@ -0,0 +1,32 @@
...
+  } else {
+    await SecureStore.deleteItemAsync(BIOMETRIC_ENABLED_KEY);
+  }
+};
+
+export const authenticate = async (): Promise<boolean> => {
+  const result = await LocalAuthentication.authenticateAsync({
+    promptMessage: "Log in to Gumroad",
+    cancelLabel: "Use password",
+    disableDeviceFallback: true,
```
pegasus1134: Intentionally no passcode fallback. If biometric fails, the user taps "Use password" which takes them to the OAuth flow instead. Mixing device passcode into the auth flow felt wrong since it's not actually authenticating against Gumroad.

PR #35, opened 2026-02-19
feat: Paginate library products (#23)

pegasus1134:
Issue: #23                                  
                                                                                                                                                                                                   
  ## Problem
                                                                                                                                                                                                   
  The library fetches all purchases in a single request with no pagination params. For users with large libraries this blocks the UI and loads excessive data upfront.                           
                                                                                                                                                                                                   
  ## Solution                                                                                                                                                                                      

  Replaced `useQuery` with `useInfiniteQuery` from React Query (already in package.json). The app now hits `mobile/purchases/index?per_page=24&page=N`, loading products in pages of 24 as the user scrolls. A loading spinner appears at the bottom while the next page fetches.

  Extracted `Purchase` type and the hook into `lib/use-purchases.ts` so both the library screen and the purchase detail page share the same paginated cache. The detail page gets a `usePurchase(id)` helper that looks up a single purchase from the flattened pages.

  All filtering, sorting, and search stay client-side on the accumulated pages — `useLibraryFilters` receives the same flat `Purchase[]` as before. Creator filter counts update progressively as pages load. This avoids depending on a new backend endpoint.

  Session expiry handling mirrors `useAPIRequest` exactly — calls `logout()` on 401 or missing token. Pull-to-refresh re-fetches all loaded pages.

  Page size of 24 gives 12 rows per page in the 2-column grid. Good balance of initial load speed vs scroll depth before the next fetch.

  ## Screenshots

  ### Before — light mode (no pagination, all products load at once)


https://github.com/user-attachments/assets/54b36f7d-6071-4b4f-88af-58b428d72f01


  ### Before — dark mode (no pagination, all products load at once)

https://github.com/user-attachments/assets/f70f08cc-6581-4e82-a84a-702c9064c353

  Page size was set to 2 for the after demos to make pagination clearly visible. Production value is 24.

  ### After — light mode (products load in pages as user scrolls)

https://github.com/user-attachments/assets/ca9c5c0b-dbac-44eb-94ef-5800baff4c53


  ### After — dark mode

https://github.com/user-attachments/assets/1d056e2f-2c2b-47d3-8a78-e11d804e2804




  ### Test results
  5 suites, 55 tests (5 new), all passing.
<img width="1792" height="997" alt="Screenshot 2026-02-19 at 2 26 30 PM" src="https://github.com/user-attachments/assets/e4ff3973-c4e3-4246-b60c-c88e6ef52656" />


  ## Checklist

  - [x] Screenshots attached

  ## AI disclosure

  Used for:
  - Analyzing the existing request infrastructure and React Query setup
  - Running test suites, type checking
  - Identifying edge cases (session expiry, deep link with empty cache)

  Implementation direction and architectural decisions were my own. Client-side filtering on accumulated pages, `useInfiniteQuery` over custom pagination, page size selection, and the decision to avoid backend dependencies. All code reviewed line-by-line before committing.

### DIFF
--- a/lib/use-purchases.ts
+++ b/lib/use-purchases.ts
@@ -1,0 +1,61 @@
+import { useInfiniteQuery } from "@tanstack/react-query";
+import { useEffect } from "react";
+import { assertDefined } from "./assert";
+import { useAuth } from "./auth-context";
+import { requestAPI, UnauthorizedError } from "./request";
+
+export interface Purchase {
+  name: string;
+  creator_name: string;
+  creator_username: string;
+  creator_profile_picture_url: string;
+  thumbnail_url: string | null;
+  url_redirect_token: string;
+  purchase_email: string;
+  purchase_id?: string;
+  is_archived?: boolean;
+  content_updated_at?: string;
+  purchased_at?: string;
+  file_data?: {
+    id: string;
+    name: string;
+    filegroup?: string;
+    streaming_url?: string;
+  }[];
+}
+
+interface PurchasesResponse {
+  success: boolean;
+  products: Purchase[];
+  user_id: string;
+}
+
+const PER_PAGE = 24;
+
+export const usePurchases = () => {
+  const { accessToken, logout, isLoading: isAuthLoading } = useAuth();
+
+  const query = useInfiniteQuery<PurchasesResponse, Error, Purchase[], string[], number>({
+    queryKey: ["purchases"],
+    queryFn: ({ pageParam }) =>
+      requestAPI<PurchasesResponse>(`mobile/purchases/index?per_page=${PER_PAGE}&page=${pageParam}`, {
+        accessToken: assertDefined(accessToken),
+      }),
+    initialPageParam: 1,
+    getNextPageParam: (lastPage, _allPages, lastPageParam) =>
+      lastPage.products.length < PER_PAGE ? undefined : lastPageParam + 1,
+    select: (data) => data.pages.flatMap((page) => page.products),
+    enabled: !!accessToken,
+  });
+
+  useEffect(() => {
+    if ((!isAuthLoading && !accessToken) || query.error instanceof UnauthorizedError) logout();
+  }, [isAuthLoading, accessToken, query.error, logout]);
+
+  return query;
+};
+
+export const usePurchase = (id: string) => {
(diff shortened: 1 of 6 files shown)


lib/use-purchases.ts
```diff
@@ -0,0 +1,61 @@
...
+  const query = useInfiniteQuery<PurchasesResponse, Error, Purchase[], string[], number>({
+    queryKey: ["purchases"],
+    queryFn: ({ pageParam }) =>
+      requestAPI<PurchasesResponse>(`mobile/purchases/index?per_page=${PER_PAGE}&page=${pageParam}`, {
+        accessToken: assertDefined(accessToken),
+      }),
+    initialPageParam: 1,
+    getNextPageParam: (lastPage, _allPages, lastPageParam) =>
+      lastPage.products.length < PER_PAGE ? undefined : lastPageParam + 1,
+    select: (data) => data.pages.flatMap((page) => page.products),
```
pegasus1134: End-of-list detection uses a simple heuristic: if the page has fewer items than PER_PAGE, we're on the last page. This avoids needing a total_count from the backend. The only edge case is when the total is an exact multiple of PER_PAGE, which triggers one extra empty fetch, but React Query handles that gracefully by returning undefined and setting hasNextPage to false.

lib/use-purchases.ts
```diff
@@ -0,0 +1,61 @@
...
+    enabled: !!accessToken,
```
pegasus1134: The select transform flattens all pages into a single Purchase[] so every consumer (library grid, filters hook, detail page lookup) gets a simple array instead of dealing with InfiniteData page structure. This keeps the pagination mechanics fully contained in this one hook.

lib/use-purchases.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  useEffect(() => {
+    if ((!isAuthLoading && !accessToken) || query.error instanceof UnauthorizedError) logout();
+  }, [isAuthLoading, accessToken, query.error, logout]);
```
pegasus1134: Same pattern as useAPIRequest in lib/request.ts for handling session expiry. If the token disappears (e.g. async storage cleared) or the API returns a 401 UnauthorizedError, we call logout() to bounce back to the login screen. The isAuthLoading guard prevents a false positive during the initial hydration window.

Issue #36, opened 2026-02-19
Update to Boxicons v3


binary-koan:
We currently use the fully open source v2 of Boxicons, but it's missing a ton of icons we want to use in Gumroad.

We should update to v3, but since the font files are no longer published we need to find an alternative approach for including the icons. They don't have plans for a React Native package short term, and React Native doesn't support the built-in `<svg>` elements they use in the `@boxicons/react` package.

What might make sense is

* Add the `@boxicons/react` package
* Add a script that converts specific icons to use react-native-svg components and writes them to a `components/icons` directory. If it's not straightforward to just write a transformer I'd be fine with a script that just tells the user's Claude Code (or similar) to do it

Open to other suggestions though.

PR #38, opened 2026-02-19
Feat : Update to Boxicons v3

TenzDelek:
Issue: #36

# Description

Updating to Boxicons v3 from v2

## Problem
We currently use the fully open source v2 of Boxicons, but it's missing a ton of icons we want to use in Gumroad. We should update to v3, but since the font files are no longer published we need to find an alternative approach for including the icons. 

## Solution
Add a script that converts specific icons to use react-native-svg components and writes them to a components/icons directory. 

---

# Before/After

### Before

https://github.com/user-attachments/assets/26fb5af6-786b-469a-93c7-36bdcb439f33

### After

https://github.com/user-attachments/assets/1b8594cd-1814-431d-9b12-070c3d074f2c

---

# Test Results

image

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)

---

# AI Disclosure

Claude Opus 4.6 used for working out the approach and for script generation.

### DIFF
--- a/app/(tabs)/_layout.tsx
+++ b/app/(tabs)/_layout.tsx
@@ -81,7 +81,7 @@ const SettingsSheet = () => {
           <Text className="mb-4 text-sm text-muted-foreground">This will log you out of your Gumroad account.</Text>
           <Button onPress={handleLogout}>
             <Text>Logout</Text>
-            <LineIcon name="log-out" size={20} className="text-primary-foreground" />
+            <LineIcon name="door-open" size={20} className="text-primary-foreground" />
           </Button>
         </View>
         <View className="border-b border-border p-4">
@@ -92,7 +92,7 @@ const SettingsSheet = () => {
           </Text>
           <Button variant="destructive" onPress={handleDeleteAccount}>
             <Text>Go to account deletion page</Text>
-            <LineIcon name="right-arrow-alt" size={20} className="text-destructive-foreground" />
+            <LineIcon name="arrow-right-stroke" size={20} className="text-destructive-foreground" />
           </Button>
         </View>
       </SheetContent>
@@ -166,7 +166,7 @@ export default function TabsLayout() {
               title: "Analytics",
               headerLeft: () => <LogoIcon />,
               headerRight: () => <LibraryHeaderRight />,
-              tabBarIcon: ({ color, size }) => <SolidIcon name="bar-chart-alt-2" size={size} color={color} />,
+              tabBarIcon: ({ color, size }) => <SolidIcon name="bar-chart-big" size={size} color={color} />,
               href: isCreator ? undefined : null,
             }}
           />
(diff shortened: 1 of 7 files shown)


app/(tabs)/_layout.tsx
```diff
@@ -81,7 +81,7 @@ const SettingsSheet = () => {
           <Text className="mb-4 text-sm text-muted-foreground">This will log you out of your Gumroad account.</Text>
           <Button onPress={handleLogout}>
             <Text>Logout</Text>
-            <LineIcon name="log-out" size={20} className="text-primary-foreground" />
```
TenzDelek: Four v2 icon ( log-out, right-arrow-alt , bar-chart-alt-2 and carousel ) are no longer found in v3. so we have replace it with most matching icon from v3.

scripts/generate-boxicons.ts
```diff
@@ -1,42 +1,129 @@
...
+    const viewBox = packMatch[2];
+    const content = packMatch[3];
+    svgElementRegex.lastIndex = 0;
+    attrRegex.lastIndex = 0;
+    const elements = parseSvgElements(content);
+    if (elements.length > 0) {
+      entry[pack] = { viewBox, elements };
+    }
+  }
+  packEntryRegex.lastIndex = 0;
```
TenzDelek: resetting the `lastindex`. Global regexes (/g flag) maintain state between exec() calls, and without resetting they'd skip matches or fail.

scripts/generate-boxicons.ts
```diff
@@ -1,42 +1,129 @@
...
 
-const cssContent = fs.readFileSync(cssPath, "utf-8");
+const iconFiles = fs
+  .readdirSync(iconsDir)
+  .filter((f) => f.endsWith(".js") && f !== "index.js");
+
+const pathsObjRegex = /const paths = \{([\s\S]*?)\};\s*\nconst /;
+const packEntryRegex =
+  /(\w+):\s*\{\s*viewBox:\s*'([^']+)',\s*content:\s*`([^`]+)`\s*\}/g;
+const svgElementRegex = /<(\w+)\s+([^>]*?)\/>/g;
```
TenzDelek: extract each `<path .../> or <rect .../>` element from the content string

scripts/generate-boxicons.ts
```diff
@@ -1,42 +1,129 @@
...
 
-  if (prefix === "bx") {
-    lineGlyphs[name] = codePoint;
-  } else if (prefix === "bxs") {
-    solidGlyphs[name] = codePoint;
-  } else if (prefix === "bxl") {
-    logoGlyphs[name] = codePoint;
+  const objMatch = pathsObjRegex.exec(source);
+  if (!objMatch) {
+    skipped++;
```
binary-koan: I think we should raise an error if we can't parse an icon rather than skipping (unless there are known cases we should skip)

binary-koan (commented):
Does including all icons cause any issues with performance or bundle size? (especially on older/slower devices)

Intuitively I'd rather mimic the react package structure and export the icons as individual tree-shakeable components

(happy to discuss here before you open a new PR)

PR #69, opened 2026-02-20
feat: Paginate library with server-side filtering (#23)

pegasus1134:
Issue: #23

## Problem

The library fetches all products in a single API call, which is slow for large libraries and can fail entirely. Creator filter counts are computed client-side from the full dataset, so they don't update when other filters (search, archived) are active.

## Solution

Switched from the `index` endpoint to the existing `search` endpoint, which already supports pagination, full-text search, archived filtering, creator filtering, and returns seller aggregation counts as metadata in the same response. No backend changes needed.

The `search` endpoint returns `{ purchases, sellers, meta: { pagination } }` where `sellers` contains creator names and purchase counts that automatically update based on the active search query and archived filter. This is the same data the web library computes client-side, but done server-side via Elasticsearch aggregations.

### Key decisions

**`usePurchases` hook** uses `useInfiniteQuery` from TanStack Query v5, hitting `GET /mobile/purchases/search` with `items=24` per page. Filter params (`q`, `seller[]`, `archived`, `order`) are included in the query key so changing any filter resets pagination and refetches from page 1.

usePurchase uses cached data as placeholder, then fetches fresh data. Returns cached library data instantly via flatMap over the query cache, then fetches fresh data from mobile/url_redirects/get_url_redirect_attributes using the url_redirect_external_id. For direct navigation without cache (expo-web), falls back to the search endpoint with the same query key as the library, warmingthe cache as a side effect.

**`useLibraryFilters` is now state-only.** It manages filter state (search text, selected creators, archived toggle, sort order) and produces an `apiFilters` object that gets passed to `usePurchases`. Search text is debounced at 300ms. No more client-side filtering or creator count computation.

**Sort options** changed from "Recently updated" / "Purchase date" to "Newest" / "Oldest", mapped to the search endpoint's `date-desc` / `date-asc` params. The `content_updated_at` sort would require adding that field to the Elasticsearch index, which can be a follow-up.

## Changes

**Modified (5):** `app/(tabs)/library.tsx`, `app/purchase/[id].tsx`, `components/library-filters.tsx`, `components/use-library-filters.ts`, `tests/components/use-library-filters.test.ts`

**Created (2):** `lib/use-purchases.ts`, `tests/lib/use-purchases.test.ts`

## Addressing feedback from #35 and #31

- Creator counts returned as metadata in the same API call, updating based on search and archived filters
- No separate endpoints file, URLs are inline
- Clean destructuring in library.tsx (`const query = usePurchases(...)` then `const { purchases, sellers, totalCount } = query`)
- Purchase detail screen reads from cached query data instead of firing a new request
- Tests mock only `@/lib/request` with a real QueryClient, no TanStack Query mocking

### Before

https://github.com/user-attachments/assets/a78ab824-221a-475f-9255-83bd90dd2d42

https://github.com/user-attachments/assets/512ee920-2755-4155-beaf-893ae71d8e1c

### After

https://github.com/user-attachments/assets/b3905691-0a4b-471e-b805-4d6fe731f379

https://github.com/user-attachments/assets/9bf304b6-2383-4106-a680-fd98fd1c94e5

### Updated: Filtering demo and centered spinner + Persisting Search and Filter Bar:

https://github.com/user-attachments/assets/5acd3985-24e4-4a54-a1eb-151c07e97359

## Test results

 5 suites, 55 tests, all passing.

Screenshot 2026-02-28 at 6 56 26 PM

## Checklist

- [x] Screenshots attached

## AI disclosure

Used for: analyzing the search endpoint response shape, exploring the existing pagination branch, running tests and type checks.

Architecture and implementation decisions were mine. All code reviewed before committing.


Darshan3690:
hi @pegasus1134  which IDE use for run this project ? .  help

lib/use-purchases.ts
```diff
@@ -0,0 +1,111 @@
...
+export const usePurchase = (id: string): Purchase | undefined => {
+  const queryClient = useQueryClient();
+  const queries = queryClient.getQueriesData<InfiniteData<SearchResponse>>({ queryKey: ["purchases"] });
+  for (const [, data] of queries) {
+    if (!data?.pages) continue;
+    for (const page of data.pages) {
+      const match = page.purchases.find((p) => p.url_redirect_token === id);
+      if (match) return match;
+    }
+  }
```
binary-koan:
```suggestion
queries.flatMap(([, data]) => data?.pages ?? []).flatMap(page => page.purchases).find((p) => p.url_redirect_token === id)
```

Maybe something like this would be nicer than the nested for loop

pegasus1134:
> Nice! Added a few comments, and could you also please add a screencast of filtering to demonstrate the creator counts updating etc?

Thank you for the review. My latest commit addresses all the comments, and I've updated the PR description with an additional video that shows the filtering and the centered loading spinner. Tests are updated as well.

pegasus1134:
Resolved merge conflict from #70 landing on main. Ready for CI @binary-koan

Issue #27, opened 2026-02-17
Render table of contents natively on the purchase page


binary-koan:
Currently for products with multiple content pages, we show a table of contents button and previous/next buttons in the page HTML:

<img width="400" height="2622" alt="Image" src="https://github.com/user-attachments/assets/6bffa6b8-d182-4f1a-8589-e91f7f545be7" />

The UX for this is suboptimal as it's non-native and you need to scroll down to see it on long pages. We should instead use a native sticky footer bar.

Will need changes to the Gumroad web app to hide the table of contents in the mobile app view, and send messages to and from the React Native web view.

anir2003:
Opened a PR for this here: https://github.com/antiwork/gumroad-mobile/pull/47

This adds a native sticky footer on the purchase screen with Previous/Next controls and a native Table of Contents sheet, and wires a WebView bridge so navigation state stays in sync while hiding the in-page HTML controls in the app view.

teredasites:
C:/Program Files/Git/attempt #27

PR #70, opened 2026-02-21
Add native table of contents navigation for multi-page content products

yashranaway:
Issue: #27 

# Description

## Problem

Products with multiple content pages show a web-based TOC button and Previous/Next buttons at the bottom of the HTML content in the WebView. This requires scrolling down on long pages to navigate between pages, which is a poor mobile UX.

## Solution

Replace the web-based navigation with a native sticky footer bar that is always visible. The footer has Previous/Next buttons, a page counter ("2 of 5"), and a TOC button that opens a sheet with the full page list.

**Mobile app changes (`gumroad-mobile`):**
- New `ContentPageNav` component: sticky footer with Previous/Next, page counter, and TOC sheet
- Updated purchase page to handle `tocData` messages from web and send `mobileAppPageChange` messages back

**Web changes (`gumroad`):**
- Hide web TOC/Previous/Next nav when in mobile app WebView
- Send `tocData` message to React Native with pages and active index
- Listen for `mobileAppPageChange` messages from React Native

**Companion web changes:** This feature requires changes to `gumroad` web repo (`WithContent.tsx`). The implementation is ready on branch [`feat/native-toc-navigation`](https://github.com/yashranaway/gumroad/tree/feat/native-toc-navigation) but I'm unable to open a PR on `antiwork/gumroad` since it's limited to collaborators. The web-side diff is small (1 file, ~40 lines)  happy to share it however is preferred.

---

# Before/After

## before
image

## after
https://github.com/user-attachments/assets/5387b7d0-edbb-4ccd-ac2b-1d39058893f5

---

# Test Results

image

---

# Checklist

---

# AI Disclosure

Claude Opus 4.6 was used to assist with implementation and test writing.

### DIFF
--- a/app/purchase/[id].tsx
+++ b/app/purchase/[id].tsx
@@ -1,5 +1,5 @@
 import { usePurchases } from "@/app/(tabs)/library";
-import { ContentPageNav } from "@/components/content-page-nav";
+import { ContentPageNav, TocPage } from "@/components/content-page-nav";
 import { MiniAudioPlayer } from "@/components/mini-audio-player";
 import { StyledWebView } from "@/components/styled";
 import { LoadingSpinner } from "@/components/ui/loading-spinner";
@@ -36,7 +36,7 @@ type ClickMessage = {
 type TocDataMessage = {
   type: "tocData";
   payload: {
-    pages: { page_id: string; title: string | null }[];
+    pages: TocPage[];
     activePageIndex: number;
   };
 };
@@ -179,12 +179,13 @@ export default function DownloadScreen() {
           <LoadingSpinner size="large" />
         </View>
       )}
+      <View className="bg-body-bg">
+        <MiniAudioPlayer />
+      </View>
       {tocPages.length > 0 && (
         <ContentPageNav pages={tocPages} activePageIndex={activePageIndex} onPageChange={handleNativePageChange} />
       )}
-      <View className="bg-body-bg" style={{ paddingBottom: bottom }}>
-        <MiniAudioPlayer />
-      </View>
+      <View style={{ paddingBottom: bottom }} />
     </Screen>
   );
 }
--- a/components/content-page-nav.tsx
+++ b/components/content-page-nav.tsx
@@ -1,13 +1,24 @@
-import { LineIcon } from "@/components/icon";
+import { LineIcon, LineIconName } from "@/components/icon";
 import { Sheet, SheetContent, SheetHeader, SheetTitle } from "@/components/ui/sheet";
 import { Text } from "@/components/ui/text";
 import { cn } from "@/lib/utils";
 import { useState } from "react";
 import { FlatList, Pressable, View } from "react-native";
 
-type TocPage = {
+export type TocPage = {
   page_id: string;
   title: string | null;
+  icon: PageIconType;
+};
+
+type PageIconType = "license-key" | "text-only" | "video-file" | "music-file" | "mixed-files";
+
+const PAGE_ICON_COMPONENTS: Record<PageIconType, LineIconName> = {
+  "license-key": "key",
+  "text-only": "file-detail",
+  "video-file": "movie-play",
+  "music-file": "music-alt",
+  "mixed-files": "file-detail",
 };
 
 export const ContentPageNav = ({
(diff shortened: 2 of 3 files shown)


components/content-page-nav.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Sheet, SheetContent, SheetHeader, SheetTitle } from "@/components/ui/sheet";
+import { Text } from "@/components/ui/text";
+import { cn } from "@/lib/utils";
+import { useState } from "react";
+import { FlatList, Pressable, View } from "react-native";
+
+type TocPage = {
+  page_id: string;
+  title: string | null;
+};
```
yashranaway: Kept this as a local type rather than importing from the purchase page since this component doesn't need to know about the message protocol — it just needs page IDs and titles.

components/content-page-nav.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const hasPrevious = activePageIndex > 0;
+  const hasNext = activePageIndex < pages.length - 1;
+
+  return (
+    <>
+      <View className="flex-row items-center border-t border-border bg-background px-4 py-2">
+        <Pressable
+          onPress={() => onPageChange(activePageIndex - 1)}
+          disabled={!hasPrevious}
+          className={cn("flex-row items-center gap-1", !hasPrevious && "opacity-50")}
```
yashranaway: Following the codebase pattern from `Button`, `Checkbox`, `RadioGroup` etc. where disabled Pressables get `opacity-50` rather than changing individual child colors.

app/purchase/[id].tsx
```diff
@@ -32,6 +33,14 @@ type ClickMessage = {
...
 };
 
+type TocDataMessage = {
+  type: "tocData";
+  payload: {
+    pages: { page_id: string; title: string | null }[];
+    activePageIndex: number;
+  };
+};
+
```
Comment (yashranaway) [inline app/purchase/[id].tsx]: Mirrors the exact shape sent by the web side's `useEffect` in `WithContent.tsx`. The `pages` array is a slimmed-down version of `RichContentPage` — we only need `page_id` and `title` for the native nav.

app/purchase/[id].tsx
```diff
@@ -49,6 +58,8 @@ const shareFile = async (uri: string) => {
 export default function DownloadScreen() {
   const { id } = useLocalSearchParams<{ id: string }>();
   const [isDownloading, setIsDownloading] = useState(false);
+  const [tocPages, setTocPages] = useState<TocDataMessage["payload"]["pages"]>([]);
```
Comment (yashranaway) [inline app/purchase/[id].tsx]: Reusing the type from the message definition to keep them in sync. Starts empty so the footer doesn't render until the web sends data.

app/purchase/[id].tsx
```diff
@@ -60,12 +71,25 @@ export default function DownloadScreen() {
...
+      webViewRef.current?.postMessage(JSON.stringify({ type: "mobileAppPageChange", payload: { pageIndex } }));
+    },
+    [],
+  );
+
   const handleMessage = async (event: WebViewMessageEvent) => {
     const data = event.nativeEvent.data;
     try {
-      const message = JSON.parse(data) as ClickMessage;
+      const message = JSON.parse(data) as ClickMessage | TocDataMessage;
```
Comment (yashranaway) [inline app/purchase/[id].tsx]: Changed from `as ClickMessage` to `as ClickMessage | TocDataMessage` so we can handle both message types. The `if (message.type === "tocData")` check runs first, before the existing click handling.

yashranaway (commented):
@binary-koan _a Could you please take a look?

binary-koan:
Looking at the diff on web, I think we should factor out a common hook for dealing with React Native messages - all the window vs document stuff, JSON parsing, etc shouldn't be duplicated

We also unfortunately need to detect the Expo app specifically rather than any mobile app, since I believe just checking `props.is_mobile_app_web_view` will hide it on the current native apps which still need it

I'll think about the design a bit as well, the ToC button doesn't obviously look clickable to me so I wonder if we can make it clearer (I'll think about what apps we can compare to, lmk if you have any good examples!)

yashranaway:
Looks great! Thanks.

Issue #26, opened 2026-02-13
Improve support for tapping bars in analytics charts


binary-koan:
Currently you can tap on the bar to see the analytics for that particular period. However, it would be much nicer if you could also tap above the bar (useful when the bar is short) and slide your finger around to select a different bar (useful when there are many narrow ones), like how the current app works.

Basically, when a finger is tapped/moved anywhere on the chart it should get the bar at that X value and select it.

yuliuyi717-ux:
I can take this bounty as a second lane.

Plan:
1) improve hit-testing so taps above short bars still map to the nearest bar at that X position,
2) add drag/scrub selection across the chart area for dense datasets,
3) preserve existing selection semantics/tooltips while tightening gesture handling,
4) validate behavior on both narrow and wide chart widths.

If still available, I will implement this in a focused PR.

PR #71, opened 2026-02-21
Add tap-anywhere and scrub gestures to analytics charts

yashranaway:
Issue: #26

# Description

## Problem

Tapping a bar in the analytics charts only works when you tap directly ON the bar. This is frustrating when bars are short (hard to hit) or narrow (many data points). There's also no way to slide between bars to quickly compare values.

## Solution

Added a `ChartGestureHandler` wrapper component using `react-native-gesture-handler` that overlays tap and horizontal pan gesture detection on all analytics bar charts. Users can now tap anywhere on the chart area (including above short bars) to select that bar, and slide horizontally to scrub across bars. Pan gesture disambiguates from vertical ScrollView scrolling via `activeOffsetX` / `failOffsetY`. Also fixed a stale index crash when switching time ranges with a bar selected.

---

# Before/After

https://github.com/user-attachments/assets/175206ee-0e34-41de-bf45-486b9959c40d

---

# Test Results

image

All tests pass locally.

---

# Checklist

---

# AI Disclosure

Claude Opus 4.6 was used to implement the gesture handler and perform code review.

### DIFF
--- a/components/analytics/analytics-bar-chart.ts
+++ b/components/analytics/analytics-bar-chart.ts
@@ -4,6 +4,12 @@ import { useCSSVariable } from "uniwind";
 
 export const CHART_LEFT_OFFSET = 36;
 
+export const getBarIndex = (x: number, barWidth: number, spacing: number, dataLength: number): number => {
+  const adjusted = x - CHART_LEFT_OFFSET;
+  const raw = Math.round(adjusted / (barWidth + spacing));
+  return Math.max(0, Math.min(dataLength - 1, raw));
+};
+
 export interface ChartDataPoint {
   value: number;
   label?: string;
--- a/components/analytics/chart-gesture-handler.tsx
+++ b/components/analytics/chart-gesture-handler.tsx
@@ -1,4 +1,4 @@
-import { CHART_LEFT_OFFSET } from "@/components/analytics/analytics-bar-chart";
+import { getBarIndex } from "@/components/analytics/analytics-bar-chart";
 import { useRef } from "react";
 import { View } from "react-native";
 import { Gesture, GestureDetector } from "react-native-gesture-handler";
@@ -24,27 +24,23 @@ export const ChartGestureHandler = ({
 }) => {
   const lastPanIndex = useRef(-1);
 
-  const getBarIndex = (x: number) => {
-    const adjusted = x - CHART_LEFT_OFFSET;
-    const raw = Math.round(adjusted / (barWidth + spacing));
-    return Math.max(0, Math.min(dataLength - 1, raw));
-  };
+  const resolveIndex = (x: number) => getBarIndex(x, barWidth, spacing, dataLength);
 
   const tap = Gesture.Tap()
     .maxDistance(TAP_MAX_DISTANCE)
-    .onEnd((e) => onTapBar(getBarIndex(e.x)))
+    .onEnd((e) => onTapBar(resolveIndex(e.x)))
     .runOnJS(true);
 
   const pan = Gesture.Pan()
     .activeOffsetX([-PAN_ACTIVE_OFFSET_X, PAN_ACTIVE_OFFSET_X])
     .failOffsetY([-PAN_FAIL_OFFSET_Y, PAN_FAIL_OFFSET_Y])
     .onStart((e) => {
-      const index = getBarIndex(e.x);
+      const index = resolveIndex(e.x);
       lastPanIndex.current = index;
       onScrubBar(index);
     })
     .onUpdate((e) => {
-      const index = getBarIndex(e.x);
+      const index = resolveIndex(e.x);
       if (index !== lastPanIndex.current) {
         lastPanIndex.current = index;
         onScrubBar(index);
--- a/tests/components/analytics/analytics-bar-chart.test.ts
+++ b/tests/components/analytics/analytics-bar-chart.test.ts
@@ -1,0 +1,92 @@
+import {
+  formatCurrency,
+  formatNumber,
+  getBarIndex,
+  CHART_LEFT_OFFSET,
+} from "@/components/analytics/analytics-bar-chart";
+
+describe("formatCurrency", () => {
+  it("formats cents as dollars", () => {
+    expect(formatCurrency(1500)).toBe("$15.00");
+  });
+
+  it("formats thousands with K suffix", () => {
+    expect(formatCurrency(150_000)).toBe("$1.5K");
+  });
+
+  it("formats millions with M suffix", () => {
+    expect(formatCurrency(150_000_000)).toBe("$1.5M");
+  });
+
+  it("formats zero", () => {
+    expect(formatCurrency(0)).toBe("$0.00");
+  });
+
+  it("returns $0.00 for null", () => {
+    expect(formatCurrency(null as unknown as number)).toBe("$0.00");
+  });
+
+  it("returns $0.00 for undefined", () => {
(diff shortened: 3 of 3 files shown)


components/analytics/chart-gesture-handler.tsx
```diff
@@ -0,0 +1,61 @@
...
+  barWidth: number;
+  spacing: number;
+  dataLength: number;
+  onTapBar: (index: number) => void;
+  onScrubBar: (index: number) => void;
+  children: React.ReactNode;
+}) => {
+  const lastPanIndex = useRef(-1);
+
+  const resolveIndex = (x: number) => getBarIndex(x, barWidth, spacing, dataLength);
```
yashranaway: Thin wrapper around the extracted `getBarIndex` that closes over the component's `barWidth`, `spacing`, and `dataLength`. Keeps the gesture callbacks clean, they just call `resolveIndex(e.x)` instead of passing four args every time.

components/analytics/chart-gesture-handler.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const tap = Gesture.Tap()
+    .maxDistance(TAP_MAX_DISTANCE)
+    .onEnd((e) => onTapBar(resolveIndex(e.x)))
+    .runOnJS(true);
```
yashranaway: Using `runOnJS(true)` because the callback just sets React state  no animation worklets or shared values involved. Running on the JS thread avoids the complexity of Reanimated interop for something this lightweight.

components/analytics/chart-gesture-handler.tsx
```diff
@@ -0,0 +1,61 @@
...
+      const index = resolveIndex(e.x);
+      if (index !== lastPanIndex.current) {
+        lastPanIndex.current = index;
+        onScrubBar(index);
+      }
+    })
+    .onEnd(() => {
+      lastPanIndex.current = -1;
+    })
+    .runOnJS(true);
```
yashranaway: The `onUpdate` callback fires on every frame during a pan, but we only want to call `onScrubBar` when the finger actually crosses into a new bar's territory. `lastPanIndex` ref tracks the previously resolved index and skips the callback if it hasn't changed. Without this, you'd get dozens of redundant state updates per swipe.

components/analytics/chart-gesture-handler.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const gesture = Gesture.Exclusive(pan, tap);
```
yashranaway: Pan gets priority over tap. If the user's finger moves horizontally past the `activeOffsetX` threshold, it becomes a pan/scrub. If they don't move (or only move vertically), the pan gesture fails and the tap fires instead. This means a quick stationary press toggles selection, and a horizontal drag scrubs  which matches how the original Gumroad app works.

components/analytics/analytics-bar-chart.ts
```diff
@@ -2,6 +2,14 @@ import { useCallback, useState } from "react";
 import { LayoutChangeEvent } from "react-native";
 import { useCSSVariable } from "uniwind";
 
+export const CHART_LEFT_OFFSET = 36;
```
yashranaway: This accounts for the ~36px of internal padding that `react-native-gifted-charts` adds on the left side for the y-axis area, even when `hideYAxisText` is set. The gesture handler needs this offset to correctly map a touch X coordinate to a bar index. The value was measured from the rendered chart if gifted-charts changes their internal layout, this would need updating.

components/analytics/analytics-bar-chart.ts
```diff
@@ -2,6 +2,14 @@ import { useCallback, useState } from "react";
...
+
+export const getBarIndex = (x: number, barWidth: number, spacing: number, dataLength: number): number => {
+  const adjusted = x - CHART_LEFT_OFFSET;
+  const raw = Math.round(adjusted / (barWidth + spacing));
+  return Math.max(0, Math.min(dataLength - 1, raw));
+};
+
```
yashranaway: Originally this lived as a closure inside `ChartGestureHandler`. Extracted it here so it's testable as a standalone pure function. Uses `Math.round` (not `Math.floor`) so tapping the right half of a gap snaps to the next bar this just feels more natural when scrubbing.

components/analytics/sales-tab.tsx
```diff
@@ -1,22 +1,25 @@
...
-  const accentColor = useCSSVariable("--color-accent") as string;
 
   const { dates, totals, sales, views } = processedData;
   const { handleLayout, barWidth, spacing } = useChartDimensions(dates.length);
 
-  const activeIndex = selectedIndex;
+  useEffect(() => {
+    setSelectedIndex(null);
+  }, [timeRange]);
+
```
yashranaway: Without this, switching from 1M (30 bars) to 1W (7 bars) while bar 25 is selected would leave `selectedIndex` pointing at an index that doesn't exist in the new data. This caused the `toLocaleString` crash hit during testing.

components/analytics/sales-tab.tsx
```diff
@@ -31,11 +34,14 @@ export const SalesTab = ({ timeRange }: { timeRange: AnalyticsTimeRange }) => {
...
+    setSelectedIndex(index);
+  }, []);
+
   const createChartData = (values: number[]) =>
     values.map((value, index) => ({
       value: value === 0 ? 0 : value,
-      frontColor: index === activeIndex ? accentColor : colors.muted,
-      onPress: () => handleBarPress(index),
+      frontColor: index === activeIndex ? colors.accent : colors.muted,
     }));
```
yashranaway: Previously each bar data item had its own `onPress: () => handleBarPress(index)`. The gesture handler now owns all touch handling for the chart area, so per-bar press handlers are no longer needed. Also replaced the standalone `useCSSVariable("--color-accent")` with `colors.accent` from `useChartColors()` which already fetches the same value was redundant.

components/analytics/sales-tab.tsx
```diff
@@ -69,25 +75,33 @@ export const SalesTab = ({ timeRange }: { timeRange: AnalyticsTimeRange }) => {
...
+                disableScroll
+                isAnimated={false}
+                barBorderTopLeftRadius={4}
+                barBorderTopRightRadius={4}
+                yAxisThickness={0}
+                xAxisThickness={1}
+                xAxisColor={colors.border}
+              />
+            </View>
+          </ChartGestureHandler>
```
yashranaway: Each BarChart gets wrapped in `ChartGestureHandler` with `barWidth`, `spacing`, and `dataLength` so the gesture handler knows the chart geometry. `onTapBar` gets the toggle callback, `onScrubBar` gets the always-set callback. Same pattern repeated for all three charts (Revenue, Sales, Views).

components/analytics/traffic-tab.tsx
```diff
@@ -35,13 +36,21 @@ export const TrafficTab = ({ timeRange }: TrafficTabProps) => {
...
   const selectedDate = activeIndex !== null && dates[activeIndex] ? dates[activeIndex] : "";
 
   const handleBarPress = useCallback((index: number) => {
     setSelectedIndex((prev) => (prev === index ? null : index));
   }, []);
 
+  const handleBarScrub = useCallback((index: number) => {
+    setSelectedIndex(index);
+  }, []);
+
```
yashranaway: Same as sales-tab. The traffic tab's charts use `stackData` instead of `data` (for the referrer color stacking), but the gesture handler doesn't care about that  it only needs the X coordinate and chart geometry.

PR #80, opened 2026-02-27
Paginate products in the library

rkluis:
Issue: #23

## Problem

The library fetches all products in a single API call via `mobile/purchases/index`, which degrades performance for large libraries and can fail entirely. Creator filter counts are computed client-side from the full dataset, so they break under pagination.

## Solution

Switched from the `index` endpoint to `mobile/purchases/search`, which supports pagination, full-text search, archived filtering, creator filtering, and returns seller aggregation counts as metadata. No backend changes needed.

### Key changes

**`usePurchases` hook** (`lib/use-purchases.ts`) uses `useInfiniteQuery` from TanStack Query v5 to hit the search endpoint with `items=24` per page. Filter params (`q`, `seller[]`, `archived`, `order`) are part of the query key, so changing any filter resets pagination and refetches from page 1. Exposes stable `refetch` and `fetchNextPage` callbacks via `useCallback` to avoid unnecessary re-renders.

**`usePurchase` hook** uses the query cache as initial data but also falls back to a search query for direct navigation (expo-web support). Uses `flatMap` chains instead of nested loops for cache lookup, per the pattern suggested in prior review feedback.

**`useLibraryFilters` is now state-only.** Manages filter UI state and produces an `apiFilters` object passed to `usePurchases`. Search text is debounced at 300ms with a named constant. No more client-side filtering or creator count computation.

**Library screen** uses `FlatList` with `onEndReached` + `onEndReachedThreshold={0.5}` for infinite scroll. Shows a centered loading spinner in the footer while fetching the next page. Error state includes a retry button.

**Sort options** changed to "Newest" / "Oldest", mapped to the search endpoint's `date-desc` / `date-asc` params.

**Sellers** come from the API response metadata, providing accurate counts that update based on the active search query and archived filter.

## Changes

**Modified (5):** `app/(tabs)/library.tsx`, `app/purchase/[id].tsx`, `components/library-filters.tsx`, `components/use-library-filters.ts`, `tests/components/use-library-filters.test.ts`

**Created (2):** `lib/use-purchases.ts`, `tests/lib/use-purchases.test.ts`

## Test results

5 suites, 56 tests, all passing. TypeScript and ESLint checks clean.

## AI disclosure

Used for: implementation of pagination hooks, updating filter logic, writing tests, and ensuring type safety.

Architecture decisions and code review were mine. All code reviewed before committing.

### DIFF
--- a/app/(tabs)/library.tsx
+++ b/app/(tabs)/library.tsx
@@ -3,56 +3,27 @@ import { LoadingSpinner } from "@/components/ui/loading-spinner";
 import { Screen } from "@/components/ui/screen";
 import { useLibraryFilters } from "@/components/use-library-filters";
 import { useAuth } from "@/lib/auth-context";
-import { useAPIRequest } from "@/lib/request";
+import { Purchase, usePurchases } from "@/lib/use-purchases";
 import { useRouter } from "expo-router";
 import { FlatList, Image, RefreshControl, Text, TouchableOpacity, View } from "react-native";
 import { useCSSVariable } from "uniwind";
 
-export interface Purchase {
-  name: string;
-  creator_name: string;
-  creator_username: string;
-  creator_profile_picture_url: string;
-  thumbnail_url: string | null;
-  url_redirect_token: string;
-  purchase_email: string;
-  purchase_id?: string;
-  is_archived?: boolean;
-  content_updated_at?: string;
-  purchased_at?: string;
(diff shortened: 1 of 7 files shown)


rkluis:
Closing this PR - upon reflection, the approach is too similar to the existing PR #69. The original author should be given priority to address the review feedback. Apologies for the noise.

PR #73, opened 2026-02-21
Improve support for tapping bars in analytics charts

dluffy56:
Issue: #26

# Description     
                                          
This PR improves the UX of analytics charts by adding tap-and-drag bar selection

## Problem

The old analytics charts only supported tapping individual bars. Users couldn't drag across bars to scrub through data and compare values.

## Solution

 Added a `ChartGestureHandler` component that places a transparent overlay over each chart and captures gestures using React Native Gesture Handler:                                   
   - Replaced the chart library's built-in `onPress` with the overlay, giving full control over touch handling                                                                           
  - Drag across bars to scrub through data in real time
  - Selection fires on touch down instead of waiting for finger lift                                                                                                                    
  - Tap a selected bar again to deselect                                                                                                                                                
  - `disablePress` prevents the chart library's internal handler from double-firing
  - Fixed stale selectedIndex when switching time ranges - if the new dataset has fewer bars than the previous one, the old index exceeds the new array bounds.
  - Added null guards to `formatCurrency` and `formatNumber` to handle undefined data

---

# Before/After

## Before
[analytics_before.webm](https://github.com/user-attachments/assets/ee4b6dff-b888-4f21-9e6e-65a6e71cf97b)

## After

https://github.com/user-attachments/assets/6ace6e70-5ada-4b5e-95e9-902bf9b011a8

---

# Test Results
- Added 19 unit tests covering index mapping logic and formatter edge cases.
- Gesture-specific e2e assertions require analytics data in the gumroad backend seed, which doesn't currently exist. The gesture logic is fully covered by the unit tests instead.

Screenshot from 2026-02-21 20-28-08

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad-mobile/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure
 Claude opus 4.6 was used to help design and review the gesture handler implementation and test coverage

PR #93, opened 2026-03-08
Add support for scrubbing and tapping above bars in the analytics charts

binary-koan:
Issue: #26 

Adds touch event handlers and an overlay component to the sales and traffic charts so that you can scrub over the chart and see the current bar even if there's no data for that period.

This performs worse than a `GestureDetector` solution unfortunately, but it's simpler and doesn't require as much fiddling with positions and offsets. 

---

# AI Disclosure

Claude Opus 4.6 used to iterate on the solution.


binary-koan:
nvm actually, https://github.com/antiwork/gumroad-mobile/pull/73 works with minimal changes and performs better so let's go with that

PR #116, opened 2026-03-22
Fix styling on tablet

binary-koan:
Issue: #112

Removes a few `sm` styles that no longer work, and fixes a couple of other bugs found while testing on tablet.


components/use-minimum-version.ts
```diff
@@ -29,7 +29,7 @@ export const useMinimumVersion = () => {
     if (!data) return null;
     return checkUpdateRequirement(
       Constants.expoConfig?.version,
-      Updates.createdAt ?? new Date("2025-03-01"),
+      Updates.createdAt ?? undefined,
```
binary-koan: Accidentally left in while testing forced updates

PR #125, opened 2026-03-29
Fix blank screen on PDF ENOENT error

gumclaw:
## Summary
- Adds error state to the PDF viewer that shows a user-facing error message with a "Try Again" button when `react-native-pdf` encounters a file error (e.g. `ENOENT` from cache eviction)
- "Try Again" clears the error and re-mounts the `Pdf` component via key counter, triggering a fresh download
- Keeps existing `Sentry.captureException` call for monitoring

Fixes https://gumroad-to.sentry.io/issues/7373029708/

## Test plan
- [ ] Verify error UI appears when PDF fails to load (covered by new unit tests)
- [ ] Verify "Try Again" re-mounts the PDF component and clears the error state
- [ ] Run `npx jest tests/app/pdf-viewer.test.tsx` — 2 tests pass
- [ ] Run `npx tsc --noEmit` — no type errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Screenshot

Simulator Screenshot - iPhone 16 - 2026-04-07 at 00 15 49

PR #126, opened 2026-03-29
Fix PDF viewer ENOENT crash after app resumes from background

gumclaw:
## Summary

- **Sentry Issue:** [GUMROAD-MOBILE-11](https://gumroad-to.sentry.io/issues/7373036293/)
- On Android, when the app goes to background and the OS cleans up cached temp files (especially under memory pressure), `react-native-pdf` tries to copy a non-existent `.pdf.tmp` file on resume, causing an unhandled ENOENT error that bypasses the `onError` prop
- Fix by listening for `AppState` changes and forcing a fresh `Pdf` component remount (via key increment) when Android returns to foreground, so the library starts a clean download instead of referencing stale temp files

## Test plan

- [x] Added tests for AppState listener registration, Android remount on resume, iOS no-op on resume, and cleanup on unmount
- [ ] Manual test: open a PDF on Android, background the app for a few minutes, return to foreground — PDF should reload cleanly
- [ ] Verify iOS behavior is unaffected

### DIFF
--- a/app/pdf-viewer.tsx
+++ b/app/pdf-viewer.tsx
@@ -61,15 +61,12 @@ export default function PdfViewerScreen() {
   }, []);
 
   const appStateRef = useRef(AppState.currentState ?? "active");
-  const handleAppStateChange = useCallback(
-    (nextAppState: string) => {
-      if (Platform.OS === "android" && appStateRef.current !== "active" && nextAppState === "active") {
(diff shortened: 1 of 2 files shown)


binary-koan:
Covered by https://github.com/antiwork/gumroad-mobile/pull/125

PR #134, opened 2026-03-31
Fix: Guard against track index out of bounds in audio player

gumclaw:
## Summary
- **`components/track-player-service.ts`**: Added bounds checking to `RemoteNext`/`RemotePrevious` lock screen handlers — these previously called `skipToNext()`/`skipToPrevious()` with no guard, causing "track index out of bounds" errors at queue edges
- **`components/use-audio-player-sync.ts`**: Added `trackIndex >= 0` guard before calling `TrackPlayer.skip()` — `findIndex` could return -1 if the track wasn't found in the list
- **`components/full-audio-player.tsx`**: Wrapped `skipToPrevious()`/`skipToNext()` in try-catch in the UI handlers — the disabled state can be stale so the native call could still throw

## Test plan
- [ ] Play audio with a single track, tap next/previous on lock screen controls — should not crash
- [ ] Play a multi-track queue, skip to the last track, tap next on lock screen — should be a no-op
- [ ] Play a multi-track queue, skip to the first track, tap previous on lock screen — should be a no-op
- [ ] Verify normal next/previous navigation still works in the full audio player UI

### DIFF
--- a/components/full-audio-player.tsx
+++ b/components/full-audio-player.tsx
@@ -95,16 +95,12 @@ export const FullAudioPlayer = ({ visible, onClose }: { visible: boolean; onClos
   };
 
   const handlePreviousTrack = async () => {
-    if (hasPrevious) {
-      await TrackPlayer.skipToPrevious();
-    }
+    if (hasPrevious) await TrackPlayer.skipToPrevious();
     await updateQueueState();
   };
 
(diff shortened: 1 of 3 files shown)


binary-koan:
@gumclaw we shouldn't catch everything, we should at least assert the error is the expected one

PR #142, opened 2026-04-01
Fix IllegalStateException: Already closed in PDF viewer

gumclaw:
## Problem

Sentry reported an `IllegalStateException: Already closed` crash in the PDF viewer on Android ([Sentry issue](https://gumroad-to.sentry.io/issues/7380320236/)).

The crash originates from `io.legere.pdfiumandroid.util.ConfigKt.handleAlreadyClosed` when the native PdfiumAndroid rendering thread tries to render a page bitmap after the PDF document has already been closed.

**Root cause:** When switching between single-page and continuous view modes, the `Pdf` component uses `key={viewMode}` which causes React to unmount the old instance and mount a new one synchronously. During unmount, the native PDF document is closed, but the background `RenderingHandler` thread may still be mid-render, leading to the `Already closed` exception.

## Fix

- Added a `pdfMounted` state that explicitly unmounts the `Pdf` component before changing the view mode
- The component is remounted on the next animation frame via `requestAnimationFrame`, giving the native rendering thread time to finish before document resources are released
- The `key={viewMode}` is preserved so React still creates a fresh instance with the new configuration

## Stack trace

```
IllegalStateException: Already closed
  RenderingHandler.handleMessage → RenderingHandler.proceed → PdfFile.renderPageBitmap
  → PdfiumCore.renderPageBitmap → PdfPage.renderPageBitmap → PdfPage.close
  → ConfigKt.handleAlreadyClosed
```

### DIFF
--- a/app/pdf-viewer.tsx
+++ b/app/pdf-viewer.tsx
@@ -136,26 +136,28 @@ export default function PdfViewerScreen() {
           ),
         }}
       />
-      {pdfMounted && <Pdf
-        key={viewMode}
-        ref={pdfRef}
-        source={{ uri }}
-        style={styles.pdf}
-        trustAllCerts={false}
-        fitPolicy={0}
-        enablePaging={viewMode === "single"}
-        horizontal={viewMode === "single"}
-        page={initialPage ? Number(initialPage) : 1}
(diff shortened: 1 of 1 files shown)


binary-koan:
Seems worth a try

PR #152, opened 2026-04-03
Fix misleading 404 when server redirects to /login

gumclaw:
## Summary
- Detect when the Gumroad API redirects unauthenticated requests to `/login` (302 → 404) and throw `UnauthorizedError` instead of the confusing "Request failed: 404 Not found" error
- Added unit tests for the `request()` function covering the redirect detection, standard 401, generic 404, and success cases

## Test plan
- [x] All 99 existing tests pass
- [x] New tests verify `UnauthorizedError` is thrown for redirected-to-login responses
- [x] New tests verify non-redirected 404s still throw the generic error

Fixes GUMROAD-MOBILE-59

### DIFF
--- a/lib/request.ts
+++ b/lib/request.ts
@@ -33,6 +33,10 @@ export const request = async <T>(
     console.info("HTTP request", details);
     throw new UnauthorizedError("Unauthorized");
   }
+  if (response.redirected && new URL(response.url).pathname === "/login") {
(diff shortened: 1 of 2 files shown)


binary-koan:
This is something we should fix in the API - it shouldn't be redirecting to login in the first place

PR #161, opened 2026-04-07
Fix: Clean up HTML error responses instead of dumping raw body into Error messages

gumclaw:
## Summary

- **Sentry issue**: [GUMROAD-MOBILE-BY](https://gumroad-to.sentry.io/issues/7392268567/) — When the Gumroad API returns a 5xx with an HTML body (e.g. Cloudflare 530 maintenance page), the `request()` function was including up to 10,000 characters of raw HTML in the `Error` message, creating massive unreadable Sentry errors.
- For HTML responses, the error message now extracts just the `<title>` text (e.g. "Maintenance | Gumroad") instead of the full HTML body.
- Adds a `ServerError` class for 5xx responses so callers can distinguish server errors from other failures (similar to the existing `UnauthorizedError`).

## Test plan

- [x] Added test: 530 with HTML body throws `ServerError` with clean title-only message
- [x] Added test: 503 with JSON body still includes the JSON text in the error
- [x] Existing tests continue to pass (7/7 green)

### DIFF
--- a/lib/request.ts
+++ b/lib/request.ts
@@ -10,6 +10,24 @@ export class UnauthorizedError extends Error {
   }
 }
 
+export class ServerError extends Error {
+  status: number;
+  constructor(status: number, message: string) {
+    super(message);
+    this.name = "ServerError";
+    this.status = status;
+  }
+}
+
+const extractErrorMessage = (body: string): string => {
+  const trimmed = body.trimStart();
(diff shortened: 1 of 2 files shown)


binary-koan:
IMO this isn't worth fixing, it's not that hard to scan for `<title>` in the Sentry UI

PR #162, opened 2026-04-07
Fix OutOfMemoryError: add response size guard

gumclaw:
## Summary

- Adds a 50 MB response size guard in `lib/request.ts` that checks the `Content-Length` header before reading the response body
- If the response exceeds the limit, the request is aborted via `AbortController` and a descriptive error is thrown
- This prevents the React Native networking bridge from base64-encoding enormous responses (the native layer encodes response bodies to bridge them to JS), which caused OOM crashes on Android

**Root cause:** React Native's networking layer calls `Base64.encodeToString` on the full response body before passing it to JavaScript. A ~163 MB response triggered a `java.lang.OutOfMemoryError: Failed to allocate a 171423608 byte allocation` in `StringFactory.newStringFromBytes`. The guard also protects the error path (`response.text()`) which previously read the entire body before slicing.

## Test plan

- [x] Added test: response with Content-Length over 50 MB throws "Response too large" error
- [x] Added test: normal-sized responses with Content-Length header still work
- [x] All existing tests pass
- [x] TypeScript compiles cleanly

### DIFF
--- a/lib/request.ts
+++ b/lib/request.ts
@@ -11,6 +11,7 @@ export class UnauthorizedError extends Error {
 }
 
 const REQUEST_TIMEOUT_MS = 30_000;
+const MAX_RESPONSE_SIZE = 50 * 1024 * 1024; // 50 MB
 
 export const request = async <T>(
   url: string,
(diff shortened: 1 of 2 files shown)


binary-koan:
I don't think this is worth following up on until there are more occurrences. We should try to work out what URL is returning so much data and fix the API rather than hacking around it

PR #166, opened 2026-04-08
Fix NSInternalInconsistencyException in PDF viewer single-page mode

gumclaw:
## Summary
- Fixes a fatal crash (`NSInternalInconsistencyException: Failed to determine navigation direction for scroll`) in the PDF viewer when users rapidly interact during single-page mode page transitions
- Adds a 350ms scroll cooldown after each page change to prevent `UIPageViewController`'s concurrent animation bug
- Sentry: https://gumroad-to.sentry.io/issues/7394880597/

## How it works
In single-page mode, `enablePaging={true}` creates a native `UIPageViewController` which crashes when a new page animation starts while one is already in-flight. The fix uses react-native-pdf's `scrollEnabled` prop to temporarily disable scroll interaction after each page transition completes, preventing rapid successive gestures from triggering concurrent animations.

## Test plan
- [x] Existing tests pass
- [x] New test: verifies scroll is disabled after page change and re-enabled after cooldown
- [ ] Manual: open a PDF in single-page mode, rapidly tap/swipe through pages — no crash

### DIFF
--- a/app/pdf-viewer.tsx
+++ b/app/pdf-viewer.tsx
@@ -9,7 +9,7 @@ import { File, Paths } from "expo-file-system";
 import * as Sharing from "expo-sharing";
 import * as Sentry from "@sentry/react-native";
 import { Stack, useLocalSearchParams } from "expo-router";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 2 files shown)


binary-koan:
I can scroll much faster than 350ms without issues, so I don't think this is the right fix - it just slows down the UX

PR #168, opened 2026-04-08
Fix SecurityException crash when third-party app denies Custom Tab launch

gumclaw:
## Summary
- Patches `expo-web-browser` native Android code to catch `SecurityException` thrown when a third-party Custom Tabs handler (e.g. OPay) registers unexported activities, preventing a crash on `launchUrl()`
- Wraps all three `launchUrl` call sites in `CustomTabsActivitiesHelper.kt` and `BrowserProxyActivity.kt` with try-catch, falling back to a plain `ACTION_VIEW` intent
- Adds `patch-package` as a devDependency to persist the native patch across installs
- Adds a JS-level test verifying the auth login flow catches browser launch errors and shows a user-friendly alert instead of crashing

## Test plan
- [ ] Verify the patch applies cleanly on fresh `npm install` (postinstall script runs `patch-package`)
- [ ] Test OAuth login flow on Android with a device that has a third-party Custom Tabs handler
- [ ] Test OAuth login flow on Android with Chrome as the default browser (happy path)
- [ ] Test OAuth login flow on iOS (unaffected, should still work)
- [ ] Confirm `npx jest` passes all tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,6 +81,7 @@
         "eslint-config-expo": "~55.0.0",
         "jest": "^29",
         "jest-expo": "~55.0.11",
+        "patch-package": "^8.0.1",
         "postcss": "^8.5.6",
         "prettier": "^3.7.4",
         "prettier-plugin-tailwindcss": "^0.7.2",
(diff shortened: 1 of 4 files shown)


binary-koan:
@gumclaw what's the sentry issue?

binary-koan:
Not enough occurrences to justify a native code patch

PR #173, opened 2026-04-09
Fix RuntimeException: Get page pdf document null

gumclaw:
## Summary
- Patches `react-native-pdf` (via `patch-package`) to catch `RuntimeException: Get page pdf document null` from the pdfium native rendering thread and route it through the `onError` callback instead of crashing the app
- Wraps `loadComplete` and `onAttachedToWindow` in try-catch to handle null document errors during page size queries and recycled view re-draws
- Installs an `UncaughtExceptionHandler` that intercepts the "pdf document null" crash from the rendering thread and dispatches it to `onError` on the main thread
- Changes `switchViewMode` in `pdf-viewer.tsx` from `requestAnimationFrame` to `setTimeout(100ms)` to give the native view time to release resources before remounting, reducing the race condition window

## Context
Sentry issue: https://gumroad-to.sentry.io/issues/7397559312/

The crash occurs in `io.legere.pdfiumandroid.PdfDocument.nativeLoadPage` when the native PDF document handle is null. This can happen when:
1. A PDF fails to initialize properly (corrupted file, incomplete download)
2. A race condition during view mode switching where the old native view's rendering thread tries to load pages after the document has been torn down

The existing `AlreadyClosedBehavior.IGNORE` config handles the "already closed" case but not the "never opened" case.

## Test plan
- [x] Added test verifying PDF unmounts immediately on view mode switch and remounts only after 100ms delay
- [x] Existing error/retry tests still pass
- [ ] Manual test: open a PDF, rapidly switch between Single Page and Continuous modes
- [ ] Manual test: open a corrupted/invalid PDF URL and verify error view appears instead of crash
- [ ] Verify on Android that the Sentry crash no longer reproduces

### DIFF
--- a/app/pdf-viewer.tsx
+++ b/app/pdf-viewer.tsx
@@ -42,10 +42,12 @@ export default function PdfViewerScreen() {
 
   const switchViewMode = (mode: "single" | "continuous") => {
     // Unmount the PDF component first to let the native rendering thread finish
-    // before the document is closed, avoiding IllegalStateException: Already closed
+    // before the document is closed, avoiding IllegalStateException: Already closed.
+    // Use a 100ms delay before remounting to give the native view time to fully
+    // release resources, preventing "Get page pdf document null" RuntimeException.
     setPdfMounted(false);
     setViewMode(mode);
-    requestAnimationFrame(() => setPdfMounted(true));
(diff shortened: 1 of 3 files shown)


binary-koan:
Not worth a native patch for 1 event, we can revisit if this comes up a lot

PR #176, opened 2026-04-09
Fix: Retry transient 502/503/504 errors in request()

gumclaw:
## Summary
- Adds retry logic with exponential backoff (500ms, 1000ms) for transient HTTP errors (502, 503, 504) in `lib/request.ts`
- Retries up to 2 times (3 total attempts) before throwing; non-transient errors (400, 401, 403, 404, 500) are not retried
- Each retry gets a fresh `AbortController` so per-attempt timeouts work correctly

**Sentry issue:** https://gumroad-to.sentry.io/issues/7398831937/
Sentry event count: 1 occurrence (first seen), minimal user impact so far.

## Test plan
- [x] Verified 502, 503, 504 are retried and succeed on retry
- [x] Verified error is thrown after all retries exhausted
- [x] Verified non-transient errors (400, 401, 500) are NOT retried
- [x] Verified exponential backoff timing (500ms then 1000ms)
- [x] All 116 existing tests pass

### DIFF
--- a/lib/request.ts
+++ b/lib/request.ts
@@ -62,7 +62,7 @@ export const request = async <T>(
       if (!response.ok) {
         if (RETRYABLE_STATUS_CODES.includes(response.status) && attempt < MAX_RETRIES) {
           console.info("HTTP request (retrying)", { ...details, attempt: attempt + 1 });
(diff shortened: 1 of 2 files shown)


binary-koan:
We should retry using react-query IMO

PR #177, opened 2026-04-09
Split submit-app into build-app and submit-app commands

slavingia:
## Summary
- Split the monolithic `/submit-app` command into two: `/build-app` (build artifacts) and `/submit-app` (submit to stores)
- Both commands now auto-fetch `.env.build.local` and `google-services.json` from 1Password so builds don't get stuck on missing credentials
- Fixed `.gitignore` `build-*` pattern that was blocking the new command file

## Test plan
- [ ] Run `/build-app` and verify it fetches credentials from 1Password
- [ ] Run `/submit-app` with pre-built artifacts
- [ ] Verify the two-step workflow works end-to-end

### DIFF
--- a/.claude/commands/build-app.md
+++ b/.claude/commands/build-app.md
@@ -53,7 +53,7 @@ Argument: platform — one of "ios", "android", or "both" (default: "both")
 
 2. **Fastlane**: Required for iOS builds (EAS uses it for signing/archiving). Install with `arch -arm64 brew install fastlane` if missing.
 
-3. **JDK 17+**: Required for Android builds (Gradle 9 needs it). Install with `arch -arm64 brew install openjdk@17` and symlink:
+3. **JDK 17+**: Required for Android builds (Gradle 9 needs it). If `$JAVA_HOME` is not already pointing to a JDK 17 install, install it with `arch -arm64 brew install openjdk@17` and symlink:
 
    ```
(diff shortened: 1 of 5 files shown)


slavingia:
Hey @binary-koan — updated the build command with learnings from running this locally for the first time:

**Prerequisites that aren't obvious:**
- **EAS CLI** needs to be installed globally (`npm install -g eas-cli`) and logged in via `eas login --sso` — the project is under the `anti-work` Expo org, so the user needs to be a member
- **Fastlane** is required even for iOS builds (EAS uses it for code signing/archiving)
- **JDK 17+** is required for Android — Gradle 9 won't run on JDK 8
- **Android SDK** is required for local Android builds (install Android Studio, run the setup wizard)

**Build gotchas:**
- Parallel iOS + Android builds can cause Android OOM — bumped Gradle JVM args to `-Xmx4096m -XX:MaxMetaspaceSize=1024m`
- `JAVA_HOME` and `ANDROID_HOME` need to be set explicitly in the build command
- The `eas` npm script uses `dotenv-flow eas --` which requires `eas` in PATH (global install)

Also split the old `/submit-app` into `/build-app` and `/submit-app`, and both now auto-fetch credentials from 1Password so nobody gets stuck on missing `.env.build.local` or `google-services.json`.

binary-koan:
Tweaked so it skips a couple of steps I didn't need to do (not sure why but I already have JDK 17 from somewhere). Also added a plugin to set the gradle memory so Claude doesn't need to do it every time.

PR #179, opened 2026-04-10
Build without EAS

binary-koan:
Switches to using XCode and Android Studio directly to build, so that people don't need Expo accounts and team access.

Also removes the `fastlane` dependency by using the Google Play API directly.


binary-koan:
@slavingia Got this working, should make things simpler!

PR #184, opened 2026-04-10
Remove Sentry user feedback and link to contact form instead

binary-koan:
Removes the Sentry user feedback form and makes the button link to open a new ticket on the Gumroad website.

It was useful during the release but it's a separate place to check, so going forward it'll be easier to just keep everything in Helper.

## Before (see https://github.com/antiwork/gumroad-mobile/pull/100)

image

## After

https://github.com/user-attachments/assets/bc5a0c78-b508-4bba-8d3e-2f109b628180


binary-koan:
@slavingia FYI - probably makes more sense now that I'm not checking Sentry. It should work to integrate the Helper API like on the Gumroad frontend if we want a native form, but this is easier to get working today

PR #207, opened 2026-04-24
Fix Background ANR caused by ExoPlayer polling in background

gumclaw:
## Problem

Sentry reported a **Background ANR** in `ExoPlayerImpl.getCurrentTimeline` ([Sentry issue](https://gumroad-to.sentry.io/issues/7438937672/)).

Three separate `setInterval` polling loops fire every 5 seconds to sync playback position:
- `track-player-service.ts` — calls `TrackPlayer.getPlaybackState()`
- `use-audio-player-sync.ts` — calls `TrackPlayer.getPlaybackState()` + `getProgress()`
- `video-player.tsx` — reads `player.currentTime` and calls `updateMediaLocation`

When the app moves to the background, these intervals continue running. The underlying ExoPlayer calls (particularly `getCurrentTimeline()`) can block on the main looper, exceeding the Android background ANR threshold.

## Fix

Added an `AppState.currentState !== "active"` guard to each polling interval so they skip execution when the app is not in the foreground. This is a minimal, safe change — media playback continues normally when the app returns to the foreground, and the track-player service's event-driven callbacks (remote play/pause/stop) still work in the background.

## Changes
- `components/track-player-service.ts` — skip polling when backgrounded
- `components/use-audio-player-sync.ts` — skip polling when backgrounded
- `app/video-player.tsx` — skip polling when backgrounded

### DIFF
--- a/app/video-player.tsx
+++ b/app/video-player.tsx
@@ -8,7 +8,7 @@ import * as Sentry from "@sentry/react-native";
 import { Stack, useLocalSearchParams } from "expo-router";
 import { useVideoPlayer, VideoView } from "expo-video";
 import { useEffect, useState } from "react";
-import { StyleSheet, View } from "react-native";
+import { AppState, StyleSheet, View } from "react-native";
 
 const fetchStreamingPlaylistUrl = async (streamingUrl: string, accessToken: string): Promise<string> =>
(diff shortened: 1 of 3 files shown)


slavingia:
@claude review

PR #221, opened 2026-04-29
Fix JSApplicationIllegalArgumentException: animated node race condition on startup

gumclaw:
## Summary

Fixes a crash on low-end Android devices where `NativeAnimatedNodesManager` throws `JSApplicationIllegalArgumentException: addAnimatedEventToView: Animated node with tag [N] does not exist` during startup navigation.

**Sentry:** https://gumroad-to.sentry.io/issues/7449002858/
**Stats:** 1 occurrence, 1 user affected, first seen 2026-04-29
**Device:** Redmi A1, Android 12
**Estimated support tickets:** 0 (fatal crash before user interaction)

## Root cause

`react-native-screens` registers `Animated.event` with `useNativeDriver: true` for `onTransitionProgress` on every native stack screen. When `app/index.tsx` renders `<Redirect>` synchronously after auth finishes loading, the screen unmounts before the native animated module finishes processing the batched `addAnimatedEventToView` command — a race condition that manifests on low-end devices.

## Changes

- **`app/index.tsx`**: Replace synchronous `<Redirect>` with `router.replace()` inside a `useEffect`, deferred by one frame via `requestAnimationFrame`. This gives the native animated module time to process its queue before the view is removed.
- **`components/ui/loading-spinner.tsx`**: Add `cancelAnimation` cleanup to the rotation `useEffect` to prevent animation leaks on unmount.
- **`jest.setup.ts`**: Add `cancelAnimation` to the reanimated mock.
- **`tests/app/index.test.tsx`**: New test suite verifying deferred navigation behavior, correct routing based on auth state, and cleanup on unmount.

### DIFF
--- a/app/index.tsx
+++ b/app/index.tsx
@@ -1,5 +1,5 @@
 import { useAuth } from "@/lib/auth-context";
-import { Redirect } from "expo-router";
+import { useRouter } from "expo-router";
 import * as SplashScreen from "expo-splash-screen";
 import { useEffect } from "react";
 
@@ -7,6 +7,7 @@ SplashScreen.preventAutoHideAsync();
 
 export default function Index() {
   const { isLoading, isAuthenticated, isCreator } = useAuth();
+  const router = useRouter();
 
   useEffect(() => {
     if (!isLoading) {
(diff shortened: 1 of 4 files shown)


gumclaw:
@claude review

PR #226, opened 2026-04-29
Fix EXC_BAD_ACCESS crash in Hermes GC writeBarrierSlow

gumclaw:
## Problem

Fatal crash (`EXC_BAD_ACCESS` / `KERN_INVALID_ADDRESS at 0x10`) in `hermes::vm::HadesGC::writeBarrierSlow` on iOS.

**Sentry:** https://gumroad-to.sentry.io/issues/7450245063/
**Events (7-day):** 1 | **Users affected:** 1
**Estimated monthly tickets:** ~0 (single occurrence so far)

## Root Cause

This is a known upstream React Native bug ([facebook/react-native#54859](https://github.com/facebook/react-native/issues/54859)). When an async void TurboModule method throws an `NSException`, the catch block in `RCTTurboModule.mm` calls `convertNSExceptionToJSError` which tries to set properties on a JS object from the wrong thread. This triggers a Hermes GC write barrier crash because the GC isn't expecting mutations from a non-JS thread.

The crash path:
```
ObjCTurboModule::performVoidMethodInvocation
→ TurboModuleConvertUtils::convertNSExceptionToJSError
→ jsi::Object::setProperty (wrong thread)
→ HiddenClass::initializeMissingPropertyMap
→ HadesGC::writeBarrierSlow → KERN_INVALID_ADDRESS
```

## Fix

Bump `react-native` from `0.83.4` → `0.83.5`, which includes the upstream fix ([a9a976a](https://github.com/facebook/react-native/commit/a9a976af89c3cd52e7f742109e24dbc6ebf09aa9) by @fabriziocucci). The fix re-throws the `NSException` instead of converting it to a `JSError` on the wrong thread.

## Changes

- `package.json`: react-native 0.83.4 → 0.83.5
- `package-lock.json`: updated lockfile

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,7 @@
         "expo-widgets": "^55.0.8",
         "react": "19.2.0",
         "react-dom": "19.2.0",
-        "react-native": "0.83.4",
+        "react-native": "0.83.5",
         "react-native-android-widget": "^0.20.1",
         "react-native-blob-util": "0.24.7",
         "react-native-context-menu-view": "^1.21.0",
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #236, opened 2026-05-06
Fix app hang in ImageView.maybeRenderLocalAsset for remote URLs

gumclaw:
## Problem

Sentry: https://gumroad-to.sentry.io/issues/7464088307/

The app hangs for 2000ms+ on the main thread in `ImageView.maybeRenderLocalAsset`. This method is called during every image reload, including for remote (http/https) URLs.

The culprit is `UIImage(named:)`, which performs a synchronous search through all asset catalogs. For remote URLs, this search always returns nil but still blocks the main thread while scanning.

**Sentry event count:** 1 occurrence (first seen, newly reported)
**Users affected:** Low volume so far (single occurrence), but affects every remote image load path
**Estimated support tickets:** ~0/month (app hang, not a crash, but degrades UX)

## Fix

Add an early return in `maybeRenderLocalAsset` when the source URL has an `http` or `https` scheme. Remote URLs can never be local assets, so we skip the expensive `UIImage(named:)` call entirely.

This is added as a second hunk to the existing `expo-image+55.0.6.patch` (which already disables progressive animated image loading to prevent a similar main-thread hang).

## Changes

- `patches/expo-image+55.0.6.patch`: Added scheme check guard to `maybeRenderLocalAsset`

### DIFF
--- a/patches/expo-image+55.0.6.patch
+++ b/patches/expo-image+55.0.6.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/expo-image/ios/ImageView.swift b/node_modules/expo-image/ios/ImageView.swift
-index adcf377..9bc3480 100644
+index adcf377..14148a3 100644
 --- a/node_modules/expo-image/ios/ImageView.swift
 +++ b/node_modules/expo-image/ios/ImageView.swift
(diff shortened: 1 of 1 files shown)


ershad:
Looks extremely rare, can skip for now.

PR #215, opened 2026-04-27
Fix Background ANR caused by video player media session staying active

gumclaw:
## Summary

- Explicitly set `staysActiveInBackground = false` on the video player to prevent the media3 session service from staying active in the background
- Add an `AppState` listener to pause the player when the app is backgrounded and resume on return (if it was playing)
- Add explicit player cleanup on unmount to ensure the player is paused when navigating away

Fixes a Background ANR caused by `androidx.media3.common.util.Util.intToStringMaxRadix` in the `ExpoVideoPlaybackService` media session service.

Sentry: https://gumroad-to.sentry.io/issues/7445669840/

## Test plan

- [x] Added `tests/app/video-player.test.tsx` with 5 tests covering:
  - `staysActiveInBackground` is set to false on player setup
  - Player is paused when AppState changes to background
  - Player resumes when AppState changes to active (if was playing)
  - Player does NOT resume if it was paused before backgrounding
  - Player is paused on unmount
- [x] All 116 existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #243, opened 2026-05-18
Fix video playback resetting to beginning when app is backgrounded

### LINKED ISSUE #239: Video playback resets to beginning when app is backgrounded
## What

When a user backgrounds the Gumroad app briefly (even ~1 minute) and returns, video content restarts from the beginning instead of resuming where they left off. This affects users who consume long-form video courses on mobile.

## Why

This is a significant UX regression for creators selling video courses. Buyers are comparing the experience unfavorably to other course platforms and leaving negative app store reviews. At least one high-value creator has escalated on behalf of their customer.

## Steps to Reproduce

1. Open a purchased video product in the Gumroad mobile app
2. Start playing a video partway through (e.g. to the 5-minute mark)
3. Switch to another app or go to the home screen
4. Wait ~1 minute
5. Return to the Gumroad app

## Expected vs Actual

- **Expected:** Video resumes from where the user left off
- **Actual:** Video restarts from 0:00

## Notes

- The user reports this started after a recent app update
- Unclear if this is iOS only, Android only, or both
- No device/OS specifics yet — we've asked the customer to provide them

## Support Tickets

- [9cbe4f61](https://help.gumroad.com/all?id=9cbe4f617bb7edac24f10cfdd5c26f66)

---
Filed by Gumclaw from a support ticket. Requested by @macartgum

nyomanjyotisa:
## What

When the user backgrounds the app while watching a video and returns later, the video sometimes restarts from `0:00` instead of resuming from where they left off.

Save the playback position when the app goes to background and seek back to it when the app becomes active. The seek is only performed if the player came back with an earlier position (more than 1s behind the saved value), so we don't overwrite a position the player legitimately retained on its own.

## Why

[#215](https://github.com/antiwork/gumroad-mobile/pull/215) intentionally set `player.staysActiveInBackground = false` to fix a Background ANR caused by the media3 session service staying active in the background. With that flag off, iOS/Android may release the underlying `AVPlayer`/`ExoPlayer` resources while the app is backgrounded; when the app returns to foreground the player can come back with `currentTime = 0`, which is what the buyer in the linked support ticket experienced. The existing `AppState` listener already remembered whether the player was playing — this PR extends it to also remember the position.

Re-enabling `staysActiveInBackground` would reintroduce the ANR fixed in #215, so the fix is in JS instead.

Fixes #239.

https://github.com/user-attachments/assets/27545015-ffa6-4f96-b112-0401a4088dd9

---

AI disclosure: Authored by Claude Opus 4.7

### DIFF
--- a/app/video-player.tsx
+++ b/app/video-player.tsx
@@ -65,15 +65,24 @@ export default function VideoPlayerScreen() {
   });
 
   const wasPlayingBeforeBackgroundRef = useRef(false);
+  const positionBeforeBackgroundRef = useRef<number | null>(null);
 
   useEffect(() => {
     const subscription = AppState.addEventListener("change", (nextState: AppStateStatus) => {
       if (nextState === "background" || nextState === "inactive") {
         wasPlayingBeforeBackgroundRef.current = player.playing;
+        positionBeforeBackgroundRef.current = player.currentTime;
         player.pause();
-      } else if (nextState === "active" && wasPlayingBeforeBackgroundRef.current) {
-        player.play();
-        wasPlayingBeforeBackgroundRef.current = false;
+      } else if (nextState === "active") {
+        const savedPosition = positionBeforeBackgroundRef.current;
+        if (savedPosition !== null && player.currentTime < savedPosition - 1) {
+          player.currentTime = savedPosition;
+        }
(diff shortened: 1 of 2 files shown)


nyomanjyotisa:
@claude review

PR #245, opened 2026-05-18
Fix repeated Cloudflare captcha challenges in purchase WebView

### LINKED ISSUE #241: Repeated "Verifying you are human" captcha prompts during app usage
## What

Users are repeatedly prompted with "Verifying you are human" captcha challenges while using the app. This happens frequently during normal usage, not just on initial login.

## Why

Constant captcha interruptions break the content consumption flow, especially for users going through video courses. This is a major source of frustration and negative app store reviews.

## Possible Root Cause

This may be a symptom of the app not persisting state when backgrounded (see #239). If the webview session/cookies are lost each time the app is backgrounded, Cloudflare would treat every return as a new unauthenticated visitor and trigger a challenge. Fixing the state persistence issue in #239 may resolve this as well.

Alternatively, this could be a standalone webview cookie/session configuration issue, or Cloudflare rules being too aggressive for the app's user agent.

## Steps to Reproduce

1. Open the Gumroad mobile app
2. Browse or play purchased content
3. Switch to another app briefly, then return
4. Observe captcha challenge

## Expected vs Actual

- **Expected:** Authenticated users are not repeatedly challenged
- **Actual:** "Verifying you are human" prompts appear frequently, interrupting content consumption

## Notes

- The user reports this started after a recent app update
- Unclear if iOS only, Android only, or both
- May require investigation into both webview session handling and Cloudflare configuration

## Support Tickets

- [9cbe4f61](https://help.gumroad.com/all?id=9cbe4f617bb7edac24f10cfdd5c26f66)

---
Filed by Gumclaw from a support ticket. Requested by @macartgum

nyomanjyotisa:
## What

Users on long video courses are repeatedly hit with "Verifying you are human" Cloudflare challenges every time they leave the app and come back, interrupting playback.

The purchase WebView was mounted with `incognito`, which on iOS maps to `WKWebsiteDataStore.nonPersistentDataStore`. That data store only keeps cookies for the lifetime of the WebView instance, so as soon as iOS kills the web content process (which happens after backgrounding or under memory pressure) the Cloudflare `cf_clearance` cookie is gone and Cloudflare re-challenges.

Switch the WebView to persistent cookies (`sharedCookiesEnabled` on iOS, `thirdPartyCookiesEnabled` on Android). Add `key={accessToken}` so the WebView is fully remounted when the signed-in user changes — together with the existing `access_token` query param in the URL, this preserves the protection from [#119](https://github.com/antiwork/gumroad-mobile/pull/119) against the WebView showing a different user's session state.

## Why

`incognito` was added in [#119](https://github.com/antiwork/gumroad-mobile/pull/119) to stop "this product belongs to another email" prompts when switching users, but it has the side effect of evicting Cloudflare clearance every time the WebView remounts. Since the purchase URL already carries the access token in the query string, the server identifies the viewer from the token rather than from a session cookie, so we don't actually need a non-persistent data store to keep sessions separate — we just need to make sure the WebView remounts on user change, which `key={accessToken}` guarantees.

Fixes #241.

---

AI disclosure: Authored by Claude Opus 4.7.

### DIFF
--- a/app/purchase/[token].tsx
+++ b/app/purchase/[token].tsx
@@ -208,12 +208,14 @@ export default function DownloadScreen() {
     <Screen>
       <Stack.Screen options={{ title: purchase?.name ?? "" }} />
       <StyledWebView
+        key={accessToken ?? "anonymous"}
         ref={webViewRef}
         source={{ uri: url }}
         className="flex-1 bg-transparent"
         webviewDebuggingEnabled
         pullToRefreshEnabled
-        incognito
+        sharedCookiesEnabled
+        thirdPartyCookiesEnabled
         mediaPlaybackRequiresUserAction={false}
         originWhitelist={["*"]}
         onShouldStartLoadWithRequest={handleShouldStartLoadWithRequest}


nyomanjyotisa:
@claude review

Issue #242, opened 2026-05-15
iOS app: "Video failed to load" (Error Code 100013) on specific product — works on web


gumclaw:
A buyer reports that videos from one specific product fail to play in the iOS app (spinning circle → "Sorry, the video player failed to load. (Error Code: 100013)"), while the same videos play fine on gumroad.com. Other video products they've purchased play fine in the app.

## Reproduction
- Purchase ID: `29F5raY1t-50_wRamrrsJA==`
- Product: "Generative Patching from Beginner to Expert with VCV Rack 2"
- Failing file: `1. Comparators - Introduction` (MP4, 108.8 MB, 4m 4s) in folder `9. Comparators`
- 68 MP4 files total, organized in folders

## Devices affected (all fail)
- iPad — iOS 26.3.1(a), Gumroad app 2026.04.29
- iPad Pro — iOS 18.7.7, Gumroad app 2026.04.26
- iPhone — iOS 18.7.7, Gumroad app 2026.04.26

## What works
- Same videos stream fine on gumroad.com
- Other video products play fine in the app on the same devices

## Likely cause
Something about these specific video files (codec, resolution, file size, or folder structure) isn't handled by the iOS player. Error code 100013 may help narrow it down.

## Helper ticket
https://help.gumroad.com/all?id=41c2b29635fc73dec92e1ccf046e3091

nyomanjyotisa:
Couldn't reproduce. Impersonated the purchaser of `29F5raY1t-50_wRamrrsJA==` and `1. Comparators - Introduction` played fine: https://drive.google.com/file/d/17C89sjxzIjq_drbtrsmxmxGgUl2KvY-E/view?usp=sharing (uploaded to drive since this repo is public and the video contains real product download page)

My guess is the video wasn't ready for mobile playback yet when the buyer first tried (still being processed in the background). The app swallows playback errors silently right now, which is why we have nothing in Sentry. Opened #246 to show an error message and log the failing file context so the next case is diagnosable.

cc @macartgum

PR #246, opened 2026-05-18
Add error UI and Sentry diagnostics for failed video playback

nyomanjyotisa:
## What

When the native video player encountered an error (the buyer in #242 sees "Error Code: 100013" for one specific 108.8 MB MP4), the app's React side did nothing about it — the loading spinner disappeared and the buyer was left looking at a black screen with no explanation, and Sentry received no breadcrumb either, so we have no way to track which files fail or why.

Subscribe to `player.statusChange`. When the player enters the `error` status:
- Render an inline error message ("This video failed to load" + the underlying `error.message` for support) instead of a black screen.
- Send `Sentry.captureMessage("Video playback failed", ...)` with `videoUrl`, `sourceUri`, `streamingUrl`, `urlRedirectId`, `productFileId`, and `purchaseId` as `extra` context.

Clear the error if the player later reaches `readyToPlay` (e.g. transient buffer recovery).

## Why

The root cause of #242 is almost certainly server-side — the file in question plays fine in HTML5 on gumroad.com but fails in AVPlayer, which is consistent with either an MP4 without `moov`-atom faststart or a missing transcoded `streaming_url`. We can't fix that from the mobile app. But we can:

1. Give the buyer a clear message instead of a frozen black screen.
2. Capture enough context in Sentry that the next time someone reports a similar failure we can immediately tell which file/purchase is affected and what AVPlayer said.

Refs #242

---

AI disclosure: Authored by Claude Opus 4.7.

### DIFF
--- a/app/video-player.tsx
+++ b/app/video-player.tsx
@@ -1,4 +1,5 @@
 import { LoadingSpinner } from "@/components/ui/loading-spinner";
+import { Text } from "@/components/ui/text";
 import { useRefToLatest } from "@/components/use-ref-to-latest";
 import { useAuth } from "@/lib/auth-context";
 import { updateMediaLocation } from "@/lib/media-location";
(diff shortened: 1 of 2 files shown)


nyomanjyotisa:
@claude review

PR #248, opened 2026-05-18
Fix mobile analytics ranges and export sales

gianfrancopiana:
Ref antiwork/gumroad/issues/5115

## What

- Capped the mobile dashboard and analytics views to match the web behavior for seller analytics.
- Added a shared export action so sellers can export all sales from mobile.
- Updated the sales search and analytics queries to use the current capped range instead of the older unlimited path.

## Why

- Mobile should reflect the same reporting limits and export flow as web so sellers see consistent data across surfaces.
- This keeps the mobile experience aligned with the current product behavior instead of exposing the older "all time" path.

## Testing

- Added and updated unit tests for the analytics range logic, purchase search, and export action.
- Ran local validation and native e2e on iPhone 17 Pro / iOS 26.5, including before/after screenshot capture.

## Screenshots

| Scenario | Button | Alert | Export WebView | Native Download
|---|---|---|---|---|
| Dashboard | image | image | image | image |
| Analytics | image | image | image | image |

---

This PR was implemented with AI assistance using gpt-5.5 xhigh.


gianfrancopiana:
@claude review

PR #218, opened 2026-04-28
Restore chaching.wav sale notification sound

gianfrancopiana:
Fixes #213

## What

Bundle `chaching.wav` into the app and wire it up so sale push notifications play the coin-drop sound again.

- Added `assets/sounds/chaching.wav` (copied from `antiwork/gumroad-ios`).
- Configured the `expo-notifications` config plugin so the file is copied into the iOS app bundle and Android `res/raw` at prebuild time.
- Registered an Android `Purchases` notification channel with `sound: "chaching.wav"`, matching the `channel_id: "Purchases"` the backend sends with FCM payloads.

iOS works without app-side channel logic — APNs just looks up the sound by filename in the bundle, so as long as the wav is bundled with the exact name the backend sends, the OS plays it.

## Why

The old native iOS app shipped with this sound. The backend still sends `sound: "chaching.wav"` (`app/models/device.rb:15`) and `channel_id: "Purchases"` (`app/mailers/contacting_creator_mailer.rb:591`) on every sale notification, but the new Expo app never bundled the file.
## Test Results

Tested locally on iOS simulator (iPhone 17 Pro, iOS 26.4) by sending a fake APNs payload with `sound: "chaching.wav"` via `xcrun simctl push`. The cha-ching played as expected.

Android channel registration is verified to compile and run; full end-to-end FCM testing will need a real sale through the local Gumroad backend (or staging).

---

This PR was implemented with AI assistance using Claude Opus 4.7.

PR #255, opened 2026-05-25
Bundle chaching.caf so iOS plays cha-ching sale sound

### LINKED ISSUE #254: iOS app: sale notification "cha-ching" sound regressed to default iOS chime
## What

The Gumroad iOS Dashboard app previously played a distinctive coin-drop ("cha-ching") notification sound when a sale came in. After a recent app update, sale push notifications now play the user's default iOS notification chime — the cha-ching custom sound is gone.

## Why this matters

The cha-ching is the single most-loved feature of the Dashboard app. Sellers use it as a passive ambient sales indicator — sound alone tells them a sale happened without opening the app.

## Repro

- Install the latest Gumroad iOS app on a logged-in seller account
- Trigger a sale
- Observe push notification — plays default iOS sound, not the historical cha-ching

## Scope

iOS specifically. Android uses per-app notification channels which support a custom tone; this issue is the iOS-side custom sound bundle / APNS payload.

## Fix direction

Ship the cha-ching as a bundled `.caf` sound and reference it in the APNS payload's `aps.sound` field. Verify it survives notification grouping.

## Note

A prior internal note said this was fixed by engineering, but the seller is still reporting it as broken on 2026-05-23, so the previous fix either didn't ship to App Store or didn't actually restore the sound.

## Related (internal)

See linked private ticket in `antiwork/gumroad-private` for reporter details.

nyomanjyotisa:
## Summary

- Bundle `chaching.caf` alongside `chaching.wav` via the `expo-notifications` plugin, matching the fix direction in #254 ("ship the cha-ching as a bundled `.caf` sound").
- Future-proofs the iOS bundle for either filename in the APNs payload's `aps.sound` field.

Fixes #254 (partial — see "What I couldn't pin down" below)

## What the backend actually sends (verified)

Read against `antiwork/gumroad@main`:

- `app/models/device.rb` — `NOTIFICATION_SOUNDS = { sale: "chaching.wav" }`
- `app/mailers/contacting_creator_mailer.rb#send_push_notification!` passes that constant as the `sound:` argument for both creator and consumer apps.
- `app/services/push_notification_service/ios.rb` puts `sound:` on the `Rpush::Apns2::Notification`, which Rpush serializes to `aps.sound` in the HTTP/2 body and sets `apns-topic: com.GRD.Gumroad` as a header.
- The `ios_spec.rb` for the consumer-app sound-passed path asserts exactly that payload shape.

So today the production APNs payload says `"sound": "chaching.wav"` — **not** `chaching.caf`. The issue's "ship as `.caf`" direction does not match the current backend payload, so this PR alone won't restore the cha-ching unless the backend is also updated to point to the new filename.

## Why bundling both is still useful

1. Apple's notification docs recommend `.caf` (Core Audio Format) for APNs sounds. Shipping it gives the backend the option to flip `NOTIFICATION_SOUNDS[:sale]` to `"chaching.caf"` without another mobile release.
2. The legacy `antiwork/gumroad-ios` app shipped `chaching.wav` (I diffed it — byte-for-byte identical to `assets/sounds/chaching.wav` here, SHA `975b31a0…`). So the existing wav cannot be the format problem on its own — but having a `.caf` available is the cheapest hedge.

## What I couldn't pin down

The legacy iOS app shipped the same wav with the same filename to the same APNs payload and the cha-ching played. The new Expo app ships the same wav and it doesn't. That points at a build/packaging difference (wav present but not actually copied into the production `.app` bundle root, file-type metadata in the regenerated `pbxproj`, etc.) rather than a backend payload change. The `expo-notifications` config plugin does add the file to the Xcode project at prebuild time on my machine, but I can't inspect a production EAS-built `.ipa` from here to confirm the file is at the bundle root with the right name.

If shipping this PR doesn't restore the sound on a TestFlight build, the next things to check are:
- Pull a production EAS `.ipa`, unzip the `.app`, confirm `chaching.wav` (and now `chaching.caf`) exist at the bundle root.
- Verify on a real device with Console.app whether `UNNotificationSound(named:)` is finding the file or returning nil.
- Coordinate with backend to switch `NOTIFICATION_SOUNDS[:sale]` to `"chaching.caf"` and test the `.caf` path.

## What changed

- `assets/sounds/chaching.caf` — converted from `chaching.wav` via `afconvert -f caff -d LEI16@44100 -c 2`. Same 2.1s of audio, Linear PCM in CAF container — valid APNs sound (well under the 30s limit).
- `app.config.ts` — added the `.caf` to the `expo-notifications` plugin's `sounds` array. At prebuild time the plugin copies both files into the iOS app bundle (and into `android/app/src/main/res/raw/`, where the wav is what `setNotificationChannelAsync` references). Android keeps using the wav via the `Purchases` channel registered in `use-push-notifications.ts` — that path is unchanged.

## Test plan

- [ ] iOS simulator: `xcrun simctl push <device> com.GRD.Gumroad payload.json` with `"sound": "chaching.wav"` → cha-ching plays (regression check on #218)
- [ ] iOS simulator: same with `"sound": "chaching.caf"` → cha-ching plays (new path is wired)
- [ ] TestFlight + real sale through `antiwork/gumroad` → does the cha-ching play? If not, follow the investigation list above.
- [ ] Android: trigger a sale → cha-ching still plays via the `Purchases` channel (no regression)


nyomanjyotisa:
Closing — the premise of this PR doesn't hold up. I verified against `antiwork/gumroad@main` that the backend's APNs payload sends `"sound": "chaching.wav"` (per `app/models/device.rb`), not `chaching.caf`. The wav in this repo is also byte-for-byte identical to what `antiwork/gumroad-ios` shipped (SHA `975b31a0…`). So bundling `.caf` can't restore the cha-ching on its own — the real root cause is elsewhere. Continuing to investigate.

Issue #293, opened 2026-07-14
Purchase content WebView still loads non-web schemes (mailto:, intent:) instead of handing them to the OS


gumclaw:
#292 fixed ERR_UNKNOWN_URL_SCHEME in the payouts/profile/sales-export WebViews, but `app/purchase/[token].tsx` (line ~91) still uses the old `!/^https?:\/\//` guard — non-web scheme links inside purchased content (mailto:, intent://) load in the WebView and error instead of routing through `safeOpenURL`. Same fix shape as #292: replace the negated http check with the WebView-internal-scheme allowlist. Found as a P3 in the fable-5 review of #292.

gumclaw:
Fixed by #294 (merged). The purchase content WebView now uses the same internal-scheme allowlist as the #292 screens; mailto:/intent:// links hand off to the OS via safeOpenURL.

Note: mobile repo — merged ≠ live. This ships with the next Expo release build. Release-QA item: on Android, tap a mailto: link inside purchased content → should open the mail app instead of an in-WebView error.

PR #294, opened 2026-07-14
Hand non-web schemes in purchase content WebView to the OS

gumclaw:
Issue: antiwork/gumroad-mobile#293

## What

Fixes the last WebView still loading non-web URL schemes internally: the purchase content screen (`app/purchase/[token].tsx`). Its navigation gate used the old negated check (`!/^https?:\/\//`), so `mailto:` / `intent://` links inside purchased content loaded in the WebView and failed with `net::ERR_UNKNOWN_URL_SCHEME` instead of opening in the OS handler app.

Same fix shape as #292: allowlist only the schemes the WebView must handle internally (`about:`, `data:`, `blob:`, `javascript:`) and hand everything else to `safeOpenURL`, which resolves the handler app and alerts gracefully when none exists.

## Why

Follow-up from the fable-5 review of #292 (P3 finding). #292 fixed the payouts/profile/sales-export screens but this screen has its own separate handler that kept the buggy guard.

## Test results

- New `tests/app/purchase-token.test.tsx` (3 tests) covering: same-page/Gumroad/`about:blank`/Cloudflare/iframe navigations stay in the WebView; external web links and `mailto:`/`intent://` schemes go to `safeOpenURL` and return `false`.
- Red-first verified: with the source change stashed, the non-web-scheme test fails on the old code; passes with the fix.
- Full suite: 40 suites / 302 tests passing.
- `tsc --noEmit` exit 0, `yarn lint` clean (one pre-existing unrelated warning in `app/login.tsx`).

## Video

No UI/layout change — this is navigation-gating logic identical to what shipped in #292. Device QA note for the next Expo release: tap a `mailto:` link inside purchased content on Android → should open the mail app instead of an in-WebView error. Flagging the demo video as outstanding-for-a-maintainer since I can't record device video headlessly.

---

This PR was written by an AI agent (Hermes/gumclaw) and reviewed before submission.

### DIFF
--- a/app/purchase/[token].tsx
+++ b/app/purchase/[token].tsx
@@ -45,6 +45,13 @@ type TocDataMessage = {
   };
 };
 
+const webViewInternalSchemes = ["about:", "data:", "blob:", "javascript:"];
+
+const isWebViewInternalUrl = (url: string) => {
+  const lower = url.toLowerCase();
+  return webViewInternalSchemes.some((scheme) => lower.startsWith(scheme));
+};
+
 const downloadUrl = (token: string, productFileId: string) =>
   buildApiUrl(`/mobile/url_redirects/download/${token}/${productFileId}`);
 
@@ -88,7 +95,7 @@ export default function DownloadScreen() {
         request.url.startsWith(env.EXPO_PUBLIC_GUMROAD_URL) ||
         request.url.startsWith("https://challenges.cloudflare.com/") ||
         request.url.startsWith("https://connect-js.stripe.com/") ||
-        !/^https?:\/\//.test(request.url)
+        isWebViewInternalUrl(request.url)
       )
         return true;
       safeOpenURL(request.url);
--- a/tests/app/purchase-token.test.tsx
+++ b/tests/app/purchase-token.test.tsx
@@ -1,0 +1,114 @@
+import { render, screen } from "@testing-library/react-native";
+
+const mockUseAuth = jest.fn();
+const mockSafeOpenURL = jest.fn();
+
+jest.mock("@/lib/auth-context", () => ({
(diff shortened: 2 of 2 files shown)


gumclaw:
@claude review once

gumclaw:
## Pre-merge adversarial review (fable-5) — READY-TO-MERGE

An adversarial reviewer ran locally against head `ecd6937` with a WebView-navigation + embedded-player checklist. No P1/P2 findings.

**Verified:**
- **Red-green isolation**: reverse-applying only the source hunk makes exactly the new non-web-scheme test fail (`mailto:` expected false, got true); full `tests/app` suite 57/57 green on the branch.
- **Behavioral deltas all correct**: `mailto:`/`tel:`/`sms:`/`intent://` now hand off to the OS (the fix); `gumroadmobile://` deep links route back into the app via Linking (an improvement); malformed/no-handler URLs land in `safeOpenURL`'s graceful alert.
- **No player regression**: audio/PDF/video flows are all native and postMessage-driven — none triggers a non-http WebView navigation. iframe video embeds stay covered by the `mainDocumentURL` escape hatch.
- **No iOS bypass**: `mainDocumentURL` is iOS-only and equals the request URL on main-frame clicks, so the scheme gate still applies; Android (where ERR_UNKNOWN_URL_SCHEME lives) always hits the gate. Check ordering byte-identical to #292's screens.
- **Consistency sweep clean**: zero remaining negated-http guards in `app/`; all four `onShouldStartLoadWithRequest` sites now use the allowlist pattern.

**P3 (follow-up, not blocking):** `webViewInternalSchemes`/`isWebViewInternalUrl` is now copy-pasted in 4 screens — worth extracting to `lib/` in a small follow-up.

PR #296, opened 2026-07-15
Patch expo/fetch to guard stream controller against late native events (fixes GUMROAD-MOBILE-ZN TypeError)

gumclaw:
## What

Adds a `patch-package` patch for `expo@55.0.9` that guards `expo/fetch`'s `FetchResponse` body-stream controller against late native events, and a regression test reproducing the crash.

Stock expo registers `didComplete` / `didFailWithError` listeners that call `controller.close()` / `controller.error()` unconditionally. The patch makes each listener check the `isControllerClosed` flag first (and set it before touching the controller), so late or duplicate native events become no-ops. This is a backport of the upstream fix ([expo/expo#44909](https://github.com/expo/expo/pull/44909)), which is only in expo's unpublished changelog and not released for SDK 55.

## Why

Sentry [GUMROAD-MOBILE-ZN](https://gumroad-to.sentry.io/issues/7613344805/) — `TypeError: The stream is not in a state that permits close`, thrown from `close(expo/virtual/streams)` on Android 12, release `2026.07.13+355`.

Root cause: the agent chat client (`lib/agent.ts`) reads the SSE body via `body.getReader()` and always calls `reader.cancel()` in a `finally` block when it stops consuming (idle-timeout abort, done-event received, error). Canceling closes the JS stream controller, but the native layer can still deliver a `didComplete` / `didFailWithError` event afterwards — expo's listener then calls `controller.close()` on the already-closed controller and throws. The error surfaces out of the event listener where we have no way to catch it in app code, so a library-level guard (the same one upstream shipped) is the correct fix rather than a workaround in `lib/agent.ts`.

The patch drops out automatically when we upgrade to an expo release containing #44909 (patch-package warns on version mismatch).

## Problem

Streaming an agent reply and canceling the reader (which our client does on every stream teardown) races with native stream completion, crashing with an uncatchable `TypeError` in `expo/fetch`.

## Solution

Backport expo/expo#44909 via `patches/expo+55.0.9.patch`: check-and-set the `isControllerClosed` flag before calling `controller.close()` / `controller.error()` in the `didComplete` / `didFailWithError` listeners.

---

# Before/After

Not a UI change — dependency patch. Terminal walkthrough (regression test failing on stock expo@55.0.9, passing with the patch, then full unit suite green):

https://raw.githubusercontent.com/gumclaw/gumroad/assets-expo-fetch-stream-close/pr-expo-fetch-stream-close-walkthrough.mp4

---

# Test Results

`tests/lib/expo-fetch-stream-close.test.ts` reproduces the cancel-then-`didComplete` sequence: 2 of 3 cases fail against unpatched expo with exactly the Sentry error, all 3 pass with the patch. Full unit suite: 41 suites / 305 tests passing.

tests passing

---

# AI Disclosure

This PR was authored by Gumclaw (Hermes agent running Claude) in response to a Sentry alert webhook: root-cause analysis from the Sentry stack trace, upstream fix identification, patch generation, and regression test authoring.

### DIFF
--- a/patches/expo+55.0.9.patch
+++ b/patches/expo+55.0.9.patch
@@ -1,0 +1,26 @@
+diff --git a/node_modules/expo/src/winter/fetch/FetchResponse.ts b/node_modules/expo/src/winter/fetch/FetchResponse.ts
+index 647d803..d02bea9 100644
+--- a/node_modules/expo/src/winter/fetch/FetchResponse.ts
++++ b/node_modules/expo/src/winter/fetch/FetchResponse.ts
+@@ -42,13 +42,19 @@ export class FetchResponse extends ConcreteNativeResponse implements Response {
+           });
+ 
+           response.addListener('didComplete', () => {
+-            controller.close();
++            if (isControllerClosed) {
++              return;
++            }
+             isControllerClosed = true;
++            controller.close();
+           });
+ 
+           response.addListener('didFailWithError', (error: string) => {
+-            controller.error(new Error(error));
++            if (isControllerClosed) {
++              return;
++            }
+             isControllerClosed = true;
++            controller.error(new Error(error));
+           });
+         },
+         async pull(controller) {
--- a/tests/lib/expo-fetch-stream-close.test.ts
+++ b/tests/lib/expo-fetch-stream-close.test.ts
@@ -1,0 +1,90 @@
+// Regression test for Sentry issue GUMROAD-MOBILE-ZN:
+// "TypeError: The stream is not in a state that permits close" raised from
+// expo/fetch's FetchResponse when native delivers `didComplete` (or
+// `didFailWithError`) after the JS consumer has already canceled the body
+// stream. Our agent streaming client (lib/agent.ts) always calls
+// `reader.cancel()` when it stops reading, so a late native completion event
+// used to call `controller.close()` on an already-closed stream controller.
+//
+// The fix (patches/expo+55.0.9.patch, backport of expo/expo#44909) makes the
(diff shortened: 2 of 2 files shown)


gumclaw:
Self-review notes:

- **Why patch `src/` and not `build/`?** The published `expo` package's `fetch` entry (`node_modules/expo/fetch.js`) requires `./src/winter/fetch/index` directly — Metro compiles the TypeScript source at bundle time, and `build/winter/fetch/` only ships `.d.ts` files for this module. So the patch on `src/winter/fetch/FetchResponse.ts` is the code that actually runs.
- **Ordering matters in the guard:** the flag is set *before* calling `controller.close()`/`controller.error()` (same as upstream #44909), so a re-entrant or duplicate event during the close call still hits the early return.
- **Swallowing `didFailWithError` after cancel is intentional:** once the consumer has canceled the reader, nobody is reading the stream, so erroring the controller has no observer — it can only throw. Matches upstream behavior.
- **Patch lifecycle:** `patch-package` is version-pinned (`expo+55.0.9.patch`) and warns on mismatch at postinstall, so this drops out visibly when we upgrade to an SDK release containing expo/expo#44909.

Adversarial review verdict (pre-merge gate): no P1/P2 findings. Checked: guard ordering vs re-entrancy, the `pull()` completed-data path (also sets the flag, so a subsequent `didComplete` no-ops), the `cancel()` path (synchronous flag set, no await gap), and that the regression test fails 2/3 cases against unpatched expo with exactly the production error message before passing with the patch.

gumclaw:
@claude review once

PR #302, opened 2026-07-20
Keep stale-blob Sentry events that carry app frames; drop only frameless leftovers

gumclaw:
## What

Follow-up to #300, addressing Greptile's review on that PR (the review landed after the merge). Scopes the stale-blob Sentry filter so it only drops events with **no app stack frames**; a `Unable to resolve data for blob` event that carries app frames now still reports.

## Why

The filter added in #300 dropped every event whose exception text contained the stale-blob marker. The intent (per that PR's description) was to drop only the frameless `onunhandledrejection` leftovers from native code — the rejections already handled and retried by `lib/request.ts`. As written, a real app-side stale body read that bypassed `lib/request.ts` would have been hidden instead of surfacing with its stack trace. This adds a `hasAppFrames` check: framed events pass through, frameless ones are dropped.

---

# Before/After

Not a UI change — Sentry event filtering only. Behavior delta:

| Event | Before (#300) | After |
|---|---|---|
| Stale blob, no frames (native leftover) | dropped | dropped |
| Stale blob, only `in_app: false` frames | dropped | dropped |
| Stale blob with app frames (bypassed `lib/request.ts`) | dropped ❌ | **reported** ✅ |

---

# Test Results

Added two cases to `sentry.test.ts`: framed stale-blob event is kept; non-app-frames-only stale-blob event is still dropped.

test results

`npx tsc --noEmit` clean for touched files · prettier + eslint clean · jest 35/35 (`sentry.test.ts` + `request.test.ts`)

---

# AI Disclosure

🤖 Generated with claude-fable-5 (Hermes agent), triggered by Greptile's post-merge review on #300. Instruction lineage: triage the bot finding → confirm it real against the merged code → scope the filter with an app-frames check + tests.


gumclaw:
Superseded by #301, which shipped the identical fix (frameless-only stale-blob filtering via an app-frames check) from a parallel session and is already merged. Closing as duplicate.

PR #304, opened 2026-07-20
Treat API redirects to /login as UnauthorizedError instead of a misleading 404

gumclaw:
## What

Requests through `lib/request.ts` that the Gumroad API answers by redirecting to the login page now throw `UnauthorizedError` instead of `RequestError: Request failed: 404 Not found`. A new `isRedirectToLogin` check runs alongside the existing 401 branch: if the response was redirected and its final URL's path is `/login` (and `/login` wasn't the requested path), the request is treated as unauthorized.

## Why

Sentry issue [GUMROAD-MOBILE-Z5](https://gumroad-to.sentry.io/issues/7609030116/) (`RequestError: Request failed: 404 Not found`) has 697 events across 594 users since July 5. The event breadcrumbs show the failing requests ending at `GET https://api.gumroad.com/login` with a 404: the API responds to an invalid/expired session with a redirect to the login page rather than a 401, `fetch` follows the redirect transparently, and `/login` doesn't exist on the API host. The app then reports a misleading 404 and — more importantly — never enters the `UnauthorizedError` branch in `useAPIRequest`, so the token-refresh-and-retry path that would recover the session never runs. Users hit a dead error state instead of a silent refresh.

Classifying the redirect as `UnauthorizedError` routes these requests through the existing refresh flow (refresh token → retry; only logging out if refresh itself fails), which both fixes the UX and stops the Sentry noise — `UnauthorizedError` is already filtered in `beforeSend`.

This was previously attempted in #152 (closed); the review note there was that the API shouldn't redirect in the first place. That's still true and worth a backend fix, but 594 affected users in two weeks on shipped app versions warrants handling it client-side too — old binaries can't be patched by a server-only fix retroactively, and the client should degrade gracefully regardless.

The check requires `response.redirected === true` and excludes direct requests to `/login`, so a genuine 404 (including one from a hypothetical `/login` API route) still reports as a 404.

## Before/After

No UI change. Walkthrough of the change and the request/auth test suites passing:

https://github.com/antiwork/gumroad-mobile/blob/fix/login-redirect-unauthorized/qa-media/pr-304-login-redirect-walkthrough.mp4?raw=true

Behavior before: unauthenticated API call → redirect to `/login` → 404 → `RequestError` surfaces to the screen and Sentry. After: same response → `UnauthorizedError` → token refresh → request retried with a fresh token.

## Test Results

New tests in `tests/lib/request.test.ts` (red without the fix — the redirected-to-`/login` case failed with `RequestError` before the change):

- redirected 404 landing on `/login` → `UnauthorizedError`
- direct request to `/login` that 404s → still the generic 404 `RequestError`
- non-redirected 404 → still the generic 404 `RequestError`

Full request/auth suites locally: `tests/lib/request.test.ts`, `tests/lib/use-api-request.test.tsx`, `tests/lib/auth-context.test.tsx`, `tests/lib/authed-request.test.tsx` — 60/60 passing (shown in the video). Full `tests/` run: 318/318. `eslint` and `prettier` clean on the touched files; `tsc --noEmit` error count unchanged vs `main` (138 pre-existing, none in touched files).

---

AI disclosure: authored by Claude Fable 5 via Hermes Agent, triggered by the Sentry alert webhook for GUMROAD-MOBILE-Z5. Instructions given: triage the Sentry event, verify recurrence (>2 distinct events), identify the root cause from the stacktrace/breadcrumbs, write a failing test, fix, and open a PR with the repo's required artifacts.


gumclaw:
Self-review notes:

- **Why check `response.redirected` rather than status alone:** a genuine 404 on a real endpoint must keep reporting as a 404. Only the redirect-followed-to-`/login` shape (confirmed in the Sentry event breadcrumbs: `GET https://api.gumroad.com/login → 404`) is reclassified.
- **Why exclude `requestedUrl` path `/login`:** if the app ever requested `/login` on the API host directly, reclassifying its 404 as unauthorized would trigger a pointless token refresh loop. The guard keeps that case a plain 404.
- **`try/catch` around `new URL(...)`:** `response.url` can be an empty string in some fetch implementations and test mocks; a malformed URL should never turn into a thrown `TypeError` masking the original HTTP failure.
- **Placement before the `!response.ok` branch, gated after `status === 401`:** keeps a single throw site for `UnauthorizedError` so `useAPIRequest`'s refresh/retry and the Sentry `beforeSend` filter (which already drops `UnauthorizedError`) apply unchanged.
- Considered checking `response.redirected` inside `buildApiUrl`-scoped calls only, but `request()` is also used for the OAuth token endpoint and web URLs; the path check is host-agnostic and safe for all callers.

gumclaw:
**Pre-merge adversarial review (fable-5) — verdict: ship. No P1/P2.**

Hunt list checked against the diff at head:

1. **False-positive reclassification** — could a legitimate response get wrongly turned into `UnauthorizedError`? Requires `response.redirected === true` AND final pathname exactly `/login` AND requested pathname ≠ `/login`. Redirect-following endpoints in the app (file/asset downloads to S3, OAuth token endpoint) never land on a `/login` path; a real 404 (`redirected: false`) is untouched — covered by test.
2. **Refresh loop risk** — if the redirect-to-login persists after a token refresh, `useAPIRequest` retries once with the new token and then propagates the error (`retry: false` for `UnauthorizedError`); if refresh itself fails it logs out. No unbounded loop.
3. **200-after-redirect-to-login case** — the check intentionally sits before the `response.ok` branch, so a 200 render of the login page now throws `UnauthorizedError` instead of a JSON `SyntaxError` from parsing login HTML. Behavior improvement, not a regression.
4. **`new URL` throw safety** — `response.url` can be `""` in some fetch impls/mocks; guarded by try/catch returning `false`, so the original HTTP error still surfaces.
5. **Sentry noise path** — `UnauthorizedError` is already dropped in `beforeSend` (`lib/sentry.ts`), so the reclassified events stop reporting without any filter change.
6. **Cross-host redirect** (api host → web host `/login`) — still matches on pathname, which is the desired semantics: any login-page landing means unauthenticated.

P3 (noted, no change needed): if the backend is ever fixed to return 401 directly (the preferred fix flagged on #152), this branch becomes dead code but harmless.

PR #308, opened 2026-07-20
Filter transient FIS_AUTH_ERROR push-token failures from Sentry

gumclaw:
## What

Adds `FIS_AUTH_ERROR` to the transient-error markers in `lib/sentry.ts` so `beforeSend` drops Android push-token registration failures caused by Firebase Installations authentication errors, instead of reporting each one to Sentry as a handled exception.

## Why

`Error: Fetching the token failed: java.util.concurrent.ExecutionException: java.io.IOException: FIS_AUTH_ERROR` is the top event source in Sentry issue 7336845703 (38k+ events, 12k+ users since March). It fires when `Notifications.getDevicePushTokenAsync()` in `components/use-push-notifications.ts` fails because Firebase Installations could not authenticate the device — a transient, device-side condition (clock skew, restored backups reusing another device's installation ID, flaky Play Services) documented by Firebase as self-resolving.

The failure is already handled correctly by the app:
- **Retry:** the token fetch runs on every authenticated app launch (`usePushNotifications` effect), so registration is retried automatically the next time the app opens.
- **UX:** the failure is caught in the `.catch` of `getExpoPushToken()` — no crash, no dead screen; the user just doesn't get a push token that session.
- **Signal:** per-occurrence capture adds no actionable information. `AUTHENTICATION_FAILED` — a real FCM misconfiguration on our side — is NOT filtered and still reports (existing regression test keeps covering this).

This mirrors the existing `SERVICE_NOT_AVAILABLE` marker added in #275 for the same token-fetch path.

## Test Results

test results

- `npx jest tests/lib/sentry.test.ts` — 32/32 passing locally:
  - new drop case: `FIS_AUTH_ERROR` token-fetch failure (failed before the fix, passes after)
  - existing keep case: `AUTHENTICATION_FAILED` token-fetch failure still reports
- `npx tsc --noEmit` clean for repo files (flexile workspace errors pre-existing, unrelated)
- `eslint` + `prettier` clean on both touched files (1 pre-existing `no-require-imports` warning in the test file, present on main)

## Before/After

Not applicable — no user-facing change. Observability-only: transient FIS auth failures stop reaching Sentry; actionable FCM auth failures still do.

## QA steps

Covered by the unit tests above (the filter is pure logic over the event shape). Post-release check: event volume on Sentry issue 7336845703 drops for releases carrying this build, while `AUTHENTICATION_FAILED` events (if any) still appear.

---

## AI disclosure

🤖 Generated with claude-fable-5 (Hermes agent). Trigger: Sentry event alert webhook for issue 7336845703. Instructions: triage the alert per the retry/noise/UX/aggregate checklist, add a failing test, fix, and open a PR.


gumclaw:
Self-review:

- `lib/sentry.ts`: single marker addition to `TRANSIENT_MARKERS`, mirroring `SERVICE_NOT_AVAILABLE` from #275 — same code path (`getDevicePushTokenAsync` in `components/use-push-notifications.ts`), same transient class. The marker matches on the full exception text, so any error chain containing `FIS_AUTH_ERROR` is dropped; that's intentional — the string only ever originates from Firebase Installations auth.
- `tests/lib/sentry.test.ts`: new drop case written first and confirmed failing before the fix (RED → GREEN). The adjacent keep case (`AUTHENTICATION_FAILED`) guards against over-filtering token-fetch failures that indicate real FCM misconfiguration.
- Considered filtering at the capture site (`use-push-notifications.ts` `.catch`) instead, but `beforeSend` keeps all Sentry noise policy in one audited place with tests, consistent with the existing filters.

gumclaw:
**Pre-merge adversarial review (fable-5) — verdict: no P1/P2 findings.**

Hunt list checked:

1. **Over-broad match risk** — `TRANSIENT_MARKERS` matches on the joined exception text of the whole chain, so any event whose chain contains `FIS_AUTH_ERROR` is dropped even if the primary error is something else. Assessed: the string is emitted only by Firebase Installations auth inside the FCM token fetch; it cannot appear as a secondary cause of an unrelated app bug. Same blast-radius profile as the existing `SERVICE_NOT_AVAILABLE` marker (#275). Not a finding.
2. **Under-filtering / sibling variants** — checked the Sentry issue's event stream for other `Fetching the token failed` causes: `SERVICE_NOT_AVAILABLE` (already filtered) and `FIS_AUTH_ERROR` (this PR) cover the observed transient set; `AUTHENTICATION_FAILED` (actionable) intentionally still reports and has a keep-test.
3. **Retry story holds** — the token fetch re-runs on every authenticated app launch (`usePushNotifications` effect), so dropping the report doesn't hide a permanently-broken registration path; a systemic FCM outage would still surface via `AUTHENTICATION_FAILED` or backend device-registration metrics.
4. **Test quality** — new drop case verified RED before the fix, GREEN after; 32/32 suite passing; no fixture mutation, no snapshot brittleness.

P3 (noted, no change needed): the marker list is growing linearly; if it gains many more entries, consider structuring per-source (token-fetch vs download vs network) for readability. Not blocking.

PR #311, opened 2026-07-20
Fix 'The track index is out of bounds' when playing audio (GUMROAD-MOBILE-1W)

gumclaw:
## Problem

Sentry [GUMROAD-MOBILE-1W](https://gumroad-to.sentry.io/issues/7376100435/) — `Error: The track index is out of bounds` — **1,529 events / 761 users**, firing since March and still occurring on the current release (2026.07.02). The alert event's breadcrumbs show the exact path: user taps an audio file in the purchase WebView → `playAudio()` → the native `TrackPlayer.skip()` throws.

Two gaps in `playAudio` (`components/use-audio-player-sync.ts`):

1. **Skip index came from the JS track list, not the native queue.** When the JS ref said "same playlist" but the native player held a different (or empty) queue — playback service restarted, or a reset happened elsewhere — `TrackPlayer.skip(jsIndex)` targeted an index that doesn't exist natively and threw. (PR #134 guarded `index >= 0`, but a *valid JS index* can still be out of bounds for the *native* queue.)
2. **Concurrent `playAudio` calls could interleave.** Two rapid taps interleave `reset`/`add`/`skip` across await boundaries, so the second call could skip into a queue the first call had just cleared.

## Solution

- In the "same playlist, different track" branch, resolve the skip target against `TrackPlayer.getQueue()` by track id. If the track isn't in the native queue, rebuild the playlist (extracted `loadPlaylist()`) instead of skipping blind.
- Serialize `playAudio` calls through a promise chain so a second tap waits for the first load to settle.

---

# Before/After

Not a UI change — audio playback control flow only. The user-visible effect is that tapping an audio track after a playback-service restart now plays it instead of surfacing a "Download Failed: The track index is out of bounds" alert.

---

# Test Results

New suite `tests/components/audio-play-skip-bounds.test.ts` (4 tests). Against the **unmodified** code, 2 of 4 fail (empty-native-queue rebuild, interleaving) — confirming the failure; with the fix all pass:

```
PASS tests/components/audio-play-skip-bounds.test.ts
PASS tests/components/audio-replay-ended-track.test.ts
PASS tests/components/track-player-service.test.ts

Test Suites: 44 passed, 44 total
Tests:       360 passed, 360 total
```

`tsc --noEmit` clean for the touched files; prettier + eslint clean.

## QA steps

1. Play an audio track from a multi-file purchase, background the app, kill/restart it (or wait for the playback service to restart)
2. Reopen the purchase and tap a *different* audio track
3. Expected: track plays; no "Download Failed" alert
4. Rapid-tap two different tracks back to back — playback lands on the second track without an error

## Verification post-release

GUMROAD-MOBILE-1W event volume tapers to ~0 on the new version's adoption curve.

---

Claude Fable 5 (Hermes agent) authored this fix autonomously from the Sentry event-alert webhook. Instructions: triage the Sentry alert, reproduce the failure in a test, fix, verify green before opening a PR.

### DIFF
--- a/components/use-audio-player-sync.ts
+++ b/components/use-audio-player-sync.ts
@@ -280,7 +280,9 @@ export const useAudioPlayerSync = (webViewRef: React.RefObject<WebView | null>) 
     return () => subscription.remove();
   }, [syncMediaLocation, updateCurrentAudioRef]);
 
-  const playAudio = useCallback(
+  const playAudioChainRef = useRef<Promise<void>>(Promise.resolve());
+
+  const performPlayAudio = useCallback(
     async ({
       resourceId,
       resumeAt,
@@ -312,12 +314,7 @@ export const useAudioPlayerSync = (webViewRef: React.RefObject<WebView | null>) 
         await syncMediaLocation(position);
       }
 
-      const isNewPlaylist =
-        !previousContext ||
-        allTracksRef.current.length !== tracks.length ||
-        !allTracksRef.current.every((t, i) => t.resourceId === tracks[i].resourceId);
-
-      if (isNewPlaylist) {
+      const loadPlaylist = async () => {
         allTracksRef.current = tracks;
         await TrackPlayer.reset();
         await TrackPlayer.add(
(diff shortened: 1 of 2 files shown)


gumclaw:
Self-review notes:

- `loadPlaylist()` is a pure extraction of the previous `isNewPlaylist` body — no behavior change in that branch.
- The queue lookup keys on `t.id === audio.resourceId`, matching how tracks are added (`id: track.resourceId`). If ids ever stop being set on add, `findIndex` returns -1 and we fall back to a full rebuild — safe, just slightly wasteful.
- The serialization chain uses `.catch(() => {})` on the *previous* promise only, so an error in call N doesn't swallow or block call N+1, and the caller of each `playAudio` still sees its own rejection.
- `performPlayAudio` keeps the original dependency array (`sendAudioPlayerInfo`, `syncMediaLocation`); the wrapper depends only on `performPlayAudio`, so hook identity behavior is unchanged for consumers.
- Old-code baseline: the new test suite fails 2/4 before the fix (empty-queue rebuild + interleaving order), passes 4/4 after.

PR #313, opened 2026-07-20
Fix WatchdogTermination: cap PDF thumbnail render size and defer generation until the sheet opens

gumclaw:
## What

Reduces peak memory usage in the PDF viewer's page-navigation sheet:

- **Native thumbnail modules (iOS + Android) now render at a capped size.** Pages were rendered at full mediaBox dimensions — on iOS additionally at the device's screen scale — so a single large-format page could allocate tens of megabytes of bitmap. Thumbnails display at roughly a third of the screen width, so both modules now scale rendering to at most 480px on the longest side (iOS also pins `UIGraphicsImageRendererFormat.scale = 1`).
- **Thumbnail generation is deferred until the sheet is first opened.** `PdfNavigationSheet` is always mounted alongside the viewer and previously kicked off a download plus a thumbnail render for *every page* of the document on mount, even when the seller never tapped the pages button.

## Why

Sentry issue [GUMROAD-MOBILE-2S](https://gumroad-to.sentry.io/issues/7379668252/) — `WatchdogTermination: The OS watchdog terminated your app, possibly because it overused RAM.` — is the app's highest-impact crash: **1,356 events / 823 users** since April, still recurring daily across every recent release. Event breadcrumbs consistently end in purchase-content flows downloading PDF attachments (`files.gumroad.com/attachments/...` followed by `mobile/media_locations`), pointing at the PDF viewer as the dominant memory hotspot. A 300-page PDF previously triggered 300 full-resolution page renders the moment the viewer opened; after this change nothing renders until the sheet opens, and each render is bounded.

## Test plan

- `npx jest tests/components/pdf-navigation-sheet.test.tsx tests/app/pdf-viewer.test.tsx` — 12 passed. New tests assert no download/thumbnail work happens while the sheet has never been opened and that work starts on first open. Red-first verified: stashing the component change fails the new deferral test.
- `npx tsc --noEmit` exit 0; prettier + eslint clean on changed files.
- Native module changes (Swift/Kotlin) can't be exercised in CI — on-device verification (open a large PDF, open the pages sheet, confirm thumbnails render) is pending for a maintainer.
- Post-merge: watch GUMROAD-MOBILE-2S event rate on the next store release.

## Before/After

No intended visual change — thumbnails render in ~120px cells, well below the new 480px cap. Video pending for a maintainer with a device (native-module change; simulator capture of memory behavior isn't meaningful).

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent) from an automated Sentry alert webhook. Instructions: triage the WatchdogTermination issue, identify the memory hotspot from event breadcrumbs, write a failing test, fix, and open a PR.

### DIFF
--- a/components/pdf-navigation-sheet.tsx
+++ b/components/pdf-navigation-sheet.tsx
@@ -92,6 +92,7 @@ export const PdfNavigationSheet = ({
 }) => {
   const hasToc = tableOfContents.length > 0;
   const [activeTab, setActiveTab] = useState<"contents" | "pages">("pages");
+  const [hasOpened, setHasOpened] = useState(open);
   const [cachedUri, setCachedUri] = useState<string | null>(null);
   const [downloadError, setDownloadError] = useState(false);
   const [containerWidth, setContainerWidth] = useState(0);
@@ -104,6 +105,10 @@ export const PdfNavigationSheet = ({
   useEffect(() => {
     if (hasToc) setActiveTab("contents");
   }, [hasToc]);
+
+  useEffect(() => {
+    if (open) setHasOpened(true);
+  }, [open]);
 
   const downloadPdf = useCallback(() => {
     let cancelled = false;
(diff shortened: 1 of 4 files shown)


gumclaw:
Self-review: the deferral uses a one-way `hasOpened` latch rather than gating on `open` directly so that thumbnails generated during a previous open survive close/reopen instead of being regenerated. The 480px cap matches ~3x the displayed thumbnail cell width, so quality headroom is preserved on high-density screens while bounding per-page bitmap allocation. Native module changes mirror each other on iOS (PDFKit render scale + format.scale=1) and Android (scaled Bitmap.createBitmap).

PR #309, opened 2026-07-20
Refresh stale download URLs and retry once on 404 instead of reporting to Sentry

gumclaw:
## What

Fixes the app's biggest Sentry issue: `FileSystem.downloadFileAsync` rejecting with "response has status: 404" — [Sentry 7336845703](https://gumroad-to.sentry.io/issues/7336845703/), 38,425 events / 12,707 users since 3/15, still firing on 2026.07.13+355. (B1 item from the 2026-07-20 Sentry sweep.)

**Root cause:** download URLs embed a `url_redirect` token that can go stale between the API response and the user's tap (app backgrounded, returned later → 404), and several call sites bypassed the central download helper entirely.

## How

1. **Centralized downloads** — all five `File.downloadFileAsync` call sites now route through `downloadFileWithRetry` (`lib/file-utils.ts`). The shared URL builder moved to `lib/download-url.ts` (was duplicated in purchase + post screens).
2. **404 → refresh URL + retry once.** The helper takes an optional `refreshUrl: () => Promise<string>` callback. On a 404 it derives a fresh URL and retries exactly once:
   - **purchase/[token]** and **post/[id]**: refresh via `fetchPurchaseDetail` (extracted from `usePurchase`'s existing queryFn) → fresh `url_redirect_token` → rebuilt download URL.
   - **pdf-viewer**: same refresh when `urlRedirectId` + `productFileId` params are present; falls back to retry-same-URL when opened without purchase context.
   - **pdf-navigation-sheet**: no purchase context available (usually receives the viewer's already-cached local file) → retry-same-URL once.
   - **sales-export**: the export URL is derived from the access token, there is no fresher URL to fetch → retry-same-URL once.
3. **Stop capturing 404s that survived refresh+retry.** The helper throws a typed `FileUnavailableError` ("This file is unavailable right now…") and logs a Sentry breadcrumb instead; call sites skip `captureException` for that error and show their existing error UI (alert / "Unable to load this PDF" + Try Again). **All non-404 failures keep their existing capture behavior unchanged** (including the pre-existing single retry for connection-drop errors; non-404 HTTP statuses still fail fast with no retry).

## QA steps

- Open a purchase, background the app long enough for the url_redirect token to rotate, return, tap a file download → download succeeds after a transparent refresh+retry (no Sentry event).
- Same flow for a post attachment download and for opening a PDF from purchase content.
- Sales export: Download CSV still works; large-export and zero-byte paths unchanged.
[...]

PR #315, opened 2026-07-20
Retry sales export download once when the file arrives empty (Error: Failed to download file)

gumclaw:
## What

Retries the all-sales CSV export download once when the downloaded file arrives empty, instead of immediately failing with "Failed to download file".

## Why

Sentry issue [7579928209](https://gumroad-to.sentry.io/issues/7579928209/) — `Error: Failed to download file` in `downloadSalesExportFile` (`app/sales-export.tsx`), 5 events / 3 users since 6/28, still firing on 2026.07.13+355.

`File.downloadFileAsync` can resolve successfully but leave a zero-byte file behind when the transfer is cut off mid-flight on a flaky mobile connection. The screen already routes through `downloadFileWithRetry` (#309), but that helper only retries when the download *rejects* — a "successful" empty download slipped past it, so the user got a dead-end error alert and each occurrence was reported to Sentry.

Now the empty-file case gets one retry (delete the empty file, re-download, re-check the first byte) before failing. If it stays empty after the retry, the existing failure path is unchanged: alert + Sentry capture, so we still see genuinely broken exports.

## Test plan

- `npx jest tests/app/sales-export.test.tsx` — 8 passed:
  - new: "retries once when the download produces an empty file and shares the successful retry" (verified RED against the pre-fix code, GREEN with the fix)
  - new: "treats a download that stays empty after the retry as a failure and deletes the file" (asserts exactly 2 download attempts, Sentry capture, temp-file cleanup)
  - all existing large-export / share / failure cases unchanged
- `prettier` + `eslint` clean on changed files (remaining warnings are pre-existing mock boilerplate)

## Before/After

Not a rendered-UI change — behavior only. Before: transient empty download → immediate "Download failed" alert. After: silent retry; alert only if the retry also produces an empty file.

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent) from an automated Sentry webhook trigger. Instructions: triage Sentry issue 7579928209, identify root cause from the stacktrace, write a failing test, fix, verify locally, open a PR.

### DIFF
--- a/app/sales-export.tsx
+++ b/app/sales-export.tsx
@@ -36,18 +36,30 @@ const SALES_CSV_FILE_NAME = "sales.csv";
 const HTML_TAG_OPEN_BYTE = "<".charCodeAt(0);
 const LARGE_EXPORT_MESSAGE = "Large exports arrive by email.";
 
+const readFirstByte = (file: File) => {
+  const handle = file.open();
+  try {
+    return handle.readBytes(1)[0];
+  } finally {
+    handle.close();
+  }
+};
+
 const downloadSalesExportFile = async (url: string) => {
   const file = new File(Paths.cache, SALES_CSV_FILE_NAME);
   // The export URL is derived from the access token, so there is no fresher URL to fetch on a
   // 404 — the helper retries the same URL once.
   await downloadFileWithRetry(url, file);
+  let firstByte = readFirstByte(file);
 
-  const handle = file.open();
-  let firstByte: number | undefined;
-  try {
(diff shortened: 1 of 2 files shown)


gumclaw:
Self-review: the retry deliberately reuses `downloadFileWithRetry` so the empty-file retry also inherits the 404-refresh handling; the deletion of the empty file before re-downloading avoids sharing a stale zero-byte cache entry if the second attempt throws. The failure path (alert + Sentry capture) is intentionally unchanged so persistent empty exports remain visible.

PR #204, opened 2026-04-23
Reduce maxBreadcrumbs to 30 to prevent main-thread app hang

gumclaw:
## Problem

Sentry's `SentryCrashScopeObserver` serializes all breadcrumbs to a JSON C-string on the main thread every time a new breadcrumb is added. With the default 100 breadcrumbs, `addEscapedString` blocks the main thread for 2000ms+, triggering an App Hanging watchdog event.

**Sentry issue:** https://gumroad-to.sentry.io/issues/7435371854/
**Occurrences (7-day):** 1
**Occurrences (14-day):** 1
**Estimated users affected/month:** Low (single occurrence so far)
**Estimated support tickets/month:** 0

## Root cause

The call chain is:
1. `SentryBreadcrumbTracker.trackApplicationNotifications` fires on a UIKit notification (main thread)
2. `SentryHub.addBreadcrumb:` → `SentryScope.addBreadcrumb:`
3. `SentryCrashScopeObserver.addSerializedBreadcrumb:` → `toJSONEncodedCString:`
4. `SentryCrashJSONCodec.encode:` iterates all breadcrumbs, calling `encodeObject` → `sentrycrashjson_addStringElement` → `addQuotedEscapedString` → `addEscapedString`

With 100 breadcrumbs (the default), this serialization takes 2000ms+ on the main thread.

This is a known performance issue in the Sentry SDK's crash scope observer path. Session replay was already disabled for a similar main-thread blocking issue (see existing comment in `sentry.ts`).

## Fix

Set `maxBreadcrumbs: 30` to keep useful debugging context while keeping JSON serialization time well under the 2-second hang threshold.

## Testing

This is a configuration-only change. 30 breadcrumbs is sufficient for debugging crashes and still provides meaningful context.

PR #317, opened 2026-07-20
Fix false iOS App Hanging reports by switching to v2 app-hang detection

gumclaw:
## What

Switches iOS app-hang detection to the v2 detector (`enableAppHangTrackingV2: true`) and disables reporting of non-fully-blocking hangs (`enableReportNonFullyBlockingAppHangs: false`) in `lib/sentry.ts`.

## Why

Sentry issue [7384299696](https://gumroad-to.sentry.io/issues/7384299696/) — `App Hanging: App hanging for at least 2000 ms.` — has accumulated 1,019 events across 360 users since April and re-fires daily. Sampling 30 recent events shows the stacks are almost entirely:

- UIKit `didEnterBackground` scene transitions (the app being suspended, not frozen)
- the Sentry SDK's own session/watchdog file writes (`SentryFileManager writeData:` → `fsync`) during backgrounding
- OS framework work off our code paths (PDFKit text selection, CoreText shaping, ColorSync, accessibility/language models)

None of the sampled events contain an app frame at the hang site. This is the known false-positive mode of the v1 AppHang detector: it flags any 2s main-run-loop stall, including stalls caused by iOS suspending the app mid-run-loop. sentry-cocoa's v2 detector (8.50+; we ship 8.58.0 via `@sentry/react-native` 7.11) cancels hangs interrupted by backgrounding and reports the actual hang duration, and the non-fully-blocking category (run loop still processing some events) is not a user-visible freeze, so it's turned off.

The React Native wrapper forwards both keys to native `SentryOptions` (verified against sentry-cocoa 8.58.0's options parser); the wrapper's TS types don't declare them yet, so they're typed locally in `lib/sentry.ts`.

Real, fully-blocking hangs in app code still report — this narrows detection, it doesn't disable it. Prior work in this series: #204 (maxBreadcrumbs), #186 (session replay off), #206/#217 (image hangs).

## Test plan

- `npx jest tests/lib/sentry.test.ts` — 34 passed, including two new assertions that `Sentry.init` receives `enableAppHangTrackingV2: true` and `enableReportNonFullyBlockingAppHangs: false`
- `npx tsc --noEmit` — no new errors vs baseline; `prettier` clean on changed files
- After release, watch issue 7384299696: event rate should drop sharply while genuine fully-blocking hangs continue to report under the v2 mechanism

## Before/After

Not a rendered-UI change — Sentry SDK configuration only. Before: every ≥2s main-run-loop stall (including OS suspensions) reported as an app hang. After: only genuine fully-blocking hangs not interrupted by backgrounding report.

**Test results:**

test results

**Walkthrough video** (config diff + local test run): [pr-317-app-hang-v2-walkthrough.mp4](https://raw.githubusercontent.com/antiwork/gumroad-mobile/fix/ios-app-hang-v2-detection/qa-media/pr-317-app-hang-v2-walkthrough.mp4)

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent) from an automated Sentry webhook trigger. Instructions: triage Sentry alert for issue 7384299696, sample event stacktraces via the Sentry API, identify root cause, write a test confirming the configuration, fix, verify locally, open a PR.


gumclaw:
Self-review:

- `lib/sentry.ts`: the two app-hang keys aren't declared in `@sentry/react-native@7.11`'s `ReactNativeOptions`, so they're typed via a local `NativeAppHangOptions` intersection rather than an `as any` cast. I verified the pass-through path end to end: the RN wrapper's `initNativeSdk` only strips `beforeSend`/`beforeBreadcrumb`/`integrations`/etc. and forwards everything else, and sentry-cocoa 8.58.0's options parser explicitly reads `enableAppHangTrackingV2` and `enableReportNonFullyBlockingAppHangs` from the dict.
- `Sentry.init({...})` became `const options = {...}; Sentry.init(options)` solely so the intersection type applies — no behavior change for the existing beforeSend filtering (all 32 pre-existing filter tests still pass).
- V2 detection is scoped to iOS by the SDK itself (Android ANR tracking is a separate mechanism, unaffected).
- Risk: v2 changes the issue fingerprint/title on Sentry ("App Hanging Fully Blocked" etc.), so the old issue 7384299696 should go quiet rather than "resolve" — worth resolving it manually once the release with this config is out.

PR #330, opened 2026-07-25
Gate every PR on an Android launch smoke test

gumclaw:
## What

Adds an Android launch smoke test that runs on every pull request and push to `main`.

The workflow builds a standalone x86_64 release APK, installs it on an API 34 emulator, and verifies that a signed-out cold start reaches the login screen without a backend. It uploads the Maestro JUnit report and recording on success or failure.

The existing signed-in Maestro flows still require a reachable seeded Gumroad backend. This PR does not expose a placeholder full-suite job that can report success without running them.

## Why

Jest cannot catch a native build that fails, an APK that lacks its JavaScript bundle, a launch crash, or a broken initial route. This gives those regressions a real Android runtime gate while keeping the check independent of repository secrets.

The smoke test runs for every pull request so changes to entrypoints, Metro config, Expo plugins, assets, and other build inputs cannot bypass it.

## Before/After

CI-only change. [Successful emulator run and recording](https://github.com/antiwork/gumroad-mobile/actions/runs/30316028896/artifacts/8672447801).

## Test Results

- `actionlint .github/workflows/e2e.yml`
- Prettier check for the workflow and smoke flow
- TypeScript check
- ESLint: zero errors, one existing warning in `app/login.tsx`
- [Jest: 48 suites, 401 tests, zero failures](https://github.com/antiwork/gumroad-mobile/actions/runs/30316028811)
- [Maestro 2.7.0: 1/1 flow passed](https://github.com/antiwork/gumroad-mobile/actions/runs/30316028896)
- Codex autoreview: clean, no actionable findings

---

**AI disclosure:** The original workflow was generated with Claude Opus 5. Review fixes were implemented with OpenAI Codex and reviewed with `gpt-5.6-sol`. Prompts: “wire maestro into CI” and “fix review findings, run autoreview, and push without force.”

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -173,7 +173,9 @@ jobs:
           emulator-options: -no-window -no-snapshot-save -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: true
           script: |
-            ./gradlew -p android assembleDebug
+            # prebuild generates the Gradle wrapper inside android/, not at the repo root.
+            chmod +x android/gradlew
+            ./android/gradlew -p android assembleDebug
             adb install -r android/app/build/outputs/apk/debug/app-debug.apk
             maestro test --format junit --output maestro-report.xml .maestro/launch-smoke.yaml
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -210,7 +210,17 @@ jobs:
           script: |
             # prebuild generates the Gradle wrapper inside android/, not at the repo root.
             chmod +x android/gradlew
-            ./android/gradlew -p android assembleDebug
+            # Build native code for the emulator's ABI only. The default in
+            # android/gradle.properties is all four (armeabi-v7a, arm64-v8a, x86, x86_64),
+            # which is right for a release build but means CMake compiles reanimated,
+            # expo-modules-core and the app three extra times for architectures this job can
+            # never run. That was 28 of the job's 45 minutes, and it left too little room for
+            # the flow itself.
+            ./android/gradlew -p android assembleDebug -PreactNativeArchitectures=x86_64
+            # Hand the emulator back the RAM and cores the build was using. The daemon
+            # otherwise stays resident for the rest of the job, and adb operations against a
+            # starved emulator can stall indefinitely rather than fail.
+            ./android/gradlew -p android --stop
             adb install -r android/app/build/outputs/apk/debug/app-debug.apk
             # Every line in this block is run as its own separate shell by
             # android-emulator-runner, so multi-line shell constructs (if/then/fi, loops,
@@ -221,7 +231,14 @@ jobs:
             # e2e:android:run script invokes it. Without -e, Maestro substitutes the literal
             # string "undefined" and the run dies with the misleading "Package undefined is
             # not installed" -- minutes in, after a clean build and install have succeeded.
-            maestro test -p android -e APP_ID="$APP_ID" --format junit --output maestro-report.xml .maestro/launch-smoke.yaml
+            #
+            # The outer `timeout` is what makes a stuck flow diagnosable. The flow's own
+            # extendedWaitUntil only bounds the wait for a view, not the adb calls before it,
+            # so a wedged emulator leaves Maestro hanging on "Clear state" until the job-level
+            # timeout cancels everything -- and a cancelled job discards the artifact upload,
+            # so you get no report and no recording, only "The operation was canceled". Failing
+            # the step ourselves keeps the evidence.
+            timeout --signal=TERM --kill-after=60s 12m maestro test -p android -e APP_ID="$APP_ID" --format junit --output maestro-report.xml .maestro/launch-smoke.yaml
 
       # A failed flow is unreadable without its recording, so upload evidence either way.
       - name: Upload the report and recordings
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -212,15 +212,15 @@ jobs:
             chmod +x android/gradlew
             ./android/gradlew -p android assembleDebug
             adb install -r android/app/build/outputs/apk/debug/app-debug.apk
+            # Every line in this block is run as its own separate shell by
+            # android-emulator-runner, so multi-line shell constructs (if/then/fi, loops,
+            # heredocs) are not possible here and variables do not carry between lines.
+            #
             # Maestro does not read APP_ID out of the surrounding shell environment; a flow
             # variable has to be handed to it with -e, which is how this repo's own
             # e2e:android:run script invokes it. Without -e, Maestro substitutes the literal
             # string "undefined" and the run dies with the misleading "Package undefined is
             # not installed" -- minutes in, after a clean build and install have succeeded.
-            if [ -z "$APP_ID" ]; then
-              echo "::error::APP_ID is empty, so Maestro would drive a package named 'undefined'."
-              exit 1
-            fi
             maestro test -p android -e APP_ID="$APP_ID" --format junit --output maestro-report.xml .maestro/launch-smoke.yaml
 
       # A failed flow is unreadable without its recording, so upload evidence either way.
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,183 @@
+name: e2e
+
+# Maestro flows drive the real app on an Android emulator. There are two jobs because the
+# flows have two very different requirements:
+#
+#   smoke  — .maestro/launch-smoke.yaml only. Installs a standalone build, launches it, and
+#            asserts the signed-out cold start lands on the login screen. It touches no
+#            backend, so it can gate every pull request. It catches the failures that matter
+#            most and are invisible to jest: a build that will not boot, a crash on launch, a
+#            broken root layout, a routing change that blanks the screen for a signed-out user.
+#
+#   suite  — the whole .maestro directory. Those flows sign in as a real seller and open real
+#            purchases, so they need a reachable Gumroad backend with seeded data (.env points
+#            at http://localhost:3000). No GitHub runner has one, so this job is manual and
+#            expects a self-hosted runner, or the Rails app booted as a service here, before it
+#            can be trusted. Wiring it to pull_request today would paint every PR red and teach
+#            everyone to ignore the check.
+on:
+  pull_request:
+    paths:
+      - ".maestro/**"
+      - "app/**"
+      - "components/**"
+      - "lib/**"
+      - "modules/**"
+      - "patches/**"
+      - "app.config.ts"
+      - "package.json"
+      - "package-lock.json"
+      - ".github/workflows/e2e.yml"
+  push:
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      run_full_suite:
+        description: "Also run the flows that need a seeded backend"
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+# Each run boots an emulator and builds the app, so overlapping runs on one branch just
+# compete for the runner. A newer push makes an in-flight run obsolete.
+concurrency:
+  group: e2e-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  smoke:
+    name: launch smoke (no backend)
+    # The Android emulator needs nested virtualisation (KVM), which the Ubuntu runners
+    # provide and the macOS runners do not.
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

gumclaw:
Good catch, and it's real — `npx expo prebuild` writes the wrapper to `android/gradlew`, so `./gradlew -p android assembleDebug` exited 127 and the smoke job never got as far as installing an APK.

Fixed in 3834940: the build step now runs `./android/gradlew -p android assembleDebug`. The install path and the app-id readback were already pointed at the generated `android/` project, so nothing else moved.

The smoke run on this push is the actual proof — if it still can't build I'll fix it here rather than merge a job that has never gone green.

gumclaw:
Checked what pointing the full suite at production would actually take, since that was authorized. Two blockers, one of them fundamental.

**The seeded accounts do not exist in production.** The flows sign in as `mobile_buyer_do_not_edit@gumroad.com` (12 of 14 flows) and `mobile_seller1_do_not_edit@gumroad.com`. Both come back absent against the live database:

```
mobile_buyer_do_not_edit@gumroad.com   => ABSENT
mobile_seller1_do_not_edit@gumroad.com => ABSENT
```

They are local-seed accounts, so there is nothing to sign in as. Creating them in production would mean real user rows with a known password committed in this repo (`.maestro/utils/login-with-password.yaml` types `password`) — a permanent credential in a public repo, on an account whose library the flows read. Not something to do quietly as part of a CI change.

**And `launch.yaml` opens the Expo dev client, not the installed app:**

```
gumroadmobile://expo-development-client/?url=http%3A%2F%2F10.0.2.2%3A8082
```

Every flow starts by running it, so the suite needs a Metro bundler on port 8082 — a dev-client build, not the standalone APK this workflow produces. Repointing the backend alone would not make these flows run; they would fail at the first step regardless of which API they were aimed at.

So the `suite` job stays behind `workflow_dispatch` with its explicit secret check. That is not me being cautious about production — it is that the flows as written cannot run against it without new production accounts and a dev-client harness.

The `smoke` job is unaffected and still gates every PR: it drives the standalone build, asserts the signed-out cold start reaches the login screen, and touches no backend at all.

Worth noting for whoever picks up the seeded-backend work: booting the Rails app as a service in this job is the safer path than pointing at production, because seeds come for free and no real data is involved. I can build that separately if it is wanted.

---

*(One process note: the first replica I queried was 6889s stale and would have returned the same ABSENT answer for a wrong reason. Re-ran against `DATABASE_REPLICA2_HOST` at 1s lag to confirm.)*

gumclaw:
## The smoke job now gives a real verdict, and the reason it was failing is structural

Three fixes since the last update, each one exposing the next. Evidence for all of it is in run artifacts, not inference.

### 1. The 28-minute build was mostly wasted work

`android/gradle.properties` carries React Native's default `reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64`. The x86_64 emulator can run exactly one of those, so CMake was compiling reanimated, expo-modules-core and the app four times over. Every heavy task appears four times in the log:

```
> Task :app:buildCMakeDebug[arm64-v8a]      18:29:29
> Task :app:buildCMakeDebug[armeabi-v7a]    18:31:31
> Task :app:buildCMakeDebug[x86]            18:33:18
> Task :app:buildCMakeDebug[x86_64]         18:35:05
BUILD SUCCESSFUL in 28m 9s
```

Building `-PreactNativeArchitectures=x86_64` took that to **12m 19s**. So the "cold cache can never warm up because the job never passes" problem I described earlier was not the real blocker — I was about to propose splitting the workflow into two jobs to work around slowness that was really 4x redundant compilation. Worth naming: the fix was a property, not a redesign.

### 2. A cancelled job throws away its own evidence

The previous run hit `timeout-minutes` mid-flow. That is a **cancellation**, not a failure, and GitHub does not let the remaining steps finish — so `Upload the report and recordings` reported success while `maestro-report.xml` did not exist yet. `if: always()` does not help; it covers failure, not the job being killed. All we got was `The operation was canceled`.

Wrapping the flow in its own `timeout` makes it fail the step normally, which keeps the artifacts:

```bash
timeout --signal=TERM --kill-after=60s 12m maestro test -p android -e APP_ID="$APP_ID" ...
```

That immediately turned a 14-minute silent hang into a verdict in 51 seconds — plus a screenshot, a view hierarchy, and full logcat.

### 3. Root cause: a debug APK cannot run this app at all

```
[Failed] launch-smoke (51s) (Assertion is false: "Sign in with Gumroad" is visible)
```

The screenshot in the artifact shows the emulator sitting on the **Expo dev-client launcher** — "Development Build", "DEVELOPMENT SERVERS", `npx expo start`, a Connect button. Not a crash, not a blank screen. logcat agrees:

```
START u0 {cmp=com.antiwork.gumroadmobile/expo.modules.devlauncher.launcher.DevLauncherActivity}
```

Two things make a debug build unusable standalone, and I checked both against the actual APKs rather than reasoning from the docs:

```
debug   app-debug.apk    → 0 matches for assets/index.android.bundle
release app-release.apk  → 1 match
```

React Native's `react { }` block defaults to `debuggableVariants = ["debug"]`, which deliberately skips bundling JS for debug builds since a developer normally has Metro running. `.env` hides this locally. On top of that `expo-dev-client` routes a debug launch into `DevLauncherActivity`.

Worth flagging for whoever reads this later: `aapt2 dump badging` reports `launchable-activity: com.gumroad.app.MainActivity` for **both** variants, so the manifest does not reveal this. The missing bundle is the reliable signal.

The job now builds `assembleRelease`, signed with a throwaway keystore generated in-job (no secret needed — nothing is published), and asserts the bundle is present so this failure mode can never again masquerade as a blank screen 15 minutes in. Verified locally: `assembleRelease -PreactNativeArchitectures=x86_64` → BUILD SUCCESSFUL, `assets/index.android.bundle` present, `lib/x86_64/` only.

### Status

A run is in flight on the release build. **I have not yet seen this job green**, and given six failures in a row each of which looked like the last one, I am not calling it done until a run reports `success`. I will report the verdict either way.

One process note: three sessions were pushing to this branch at once, and `concurrency: cancel-in-progress` meant our runs kept cancelling each other's — one `cancelled` I initially read as a timeout was actually a sibling's push. Not a code problem, but it cost a diagnostic cycle.

---

Written by Hermes Agent (claude-opus-5). Instructions: diagnose the cancelled/failing e2e run on this PR, fix the root cause, verify with real CI output rather than assumption, and record the pitfalls in the team skill library.

gumclaw:
## Run 30171537423 failed for a reason the release-build switch introduced: Sentry

The flow never ran. The Gradle build died 11 minutes in, and it was not the variant, the ABI, the keystore, or Maestro — all of those worked. Steps 1 through 14 were green, including the throwaway keystore.

```
error: An organization ID or slug is required (provide with --org)
> Task :app:createBundleReleaseJsAndAssets_SentryUpload_com.antiwork.gumroadmobile@2026.07.20+1_1 FAILED
Script '…/node_modules/@sentry/react-native/sentry.gradle' line: 132
BUILD FAILED in 11m 1s
```

A debug build never does this. A release build uploads its JavaScript source maps to Sentry, and `sentry-cli` exiting non-zero fails the entire Gradle build. So this was a direct consequence of switching to `assembleRelease` in `c955b97` — a new failure mode, not a leftover.

### Why it can never succeed in this job

`android/` is gitignored, so `expo prebuild --clean` regenerates `android/sentry.properties` from the `@sentry/react-native/expo` config plugin, which fills in the organization from `SENTRY_ORG` and `SENTRY_PROJECT` (`app.config.ts` lines 134-138). This job deliberately sets no secrets so it can run on forks and every pull request, so those are unset, and the regenerated file has no `defaults.org`. sentry-cli then refuses to run.

Worth flagging for anyone reading the repo: the `android/sentry.properties` on your machine **does** name the organization (`defaults.org=gumroad-to`), which makes this look impossible. That file is gitignored and is not what CI gets.

### Why nobody would catch this locally

The gitignored `.env` carries `SENTRY_ALLOW_FAILURE`, with a comment already explaining the intent: "We don't upload source maps locally, and the lack of Sentry credentials can cause the iOS dev build to fail." That variable downgrades the exact same error to a warning:

```
error: An organization ID or slug is required (provide with --org)
Command failed, however, "SENTRY_ALLOW_FAILURE" variable or "allow-failure" flag was set. Exiting with 0 exit code.
BUILD SUCCESSFUL in 11s
```

This is the same shape as the bundle-identifier failure earlier in this PR: a value `.env` supplies locally and CI does not. That is now the second time it has cost a full run, so it is written into the team skill as a class of bug to check for rather than a one-off.

### The fix, verified locally before pushing

`eb56897` sets `SENTRY_DISABLE_AUTO_UPLOAD: "true"` on the emulator step. `sentry.gradle` gates the upload with `onlyIf { shouldSentryAutoUploadGeneral() }`, so the task is skipped outright rather than run and forgiven. Nothing about a throwaway smoke build belongs in Sentry, and the JS bundle is still built and embedded, which is all the flow needs.

I reproduced CI on my machine rather than pushing and hoping — stripping **both** local crutches (org-less `sentry.properties` as prebuild would write it, and `SENTRY_ALLOW_FAILURE` removed from `.env`), then running the same control twice:

| | result |
|---|---|
| control, no flag | `SentryUpload_… FAILED`, `An organization ID or slug is required`, **BUILD FAILED in 2m 30s** |
| `SENTRY_DISABLE_AUTO_UPLOAD=true` | `SentryUpload_… SKIPPED`, **BUILD SUCCESSFUL in 1m 58s** |

and confirmed the bundle survives the skip: `index.android.bundle`, 5,046,892 bytes. One useful detail for that loop — `--rerun-tasks` is required, because the upload task is `up-to-date` on a warm tree and a control that "passes" may have simply skipped everything.

Run **30172387595** is in flight on `eb56897`. As before, I am not calling this done until a run reports `success`; I will report the verdict either way.

---

Written by Hermes Agent (claude-opus-5). Instructions: diagnose the failing e2e run on this PR, fix the root cause, verify with real execution rather than assumption, and record the pitfall in the team skill library.

gumclaw:
## Green. The smoke job passes end to end, and the last failure was the most interesting one

Run [30173157876](https://github.com/antiwork/gumroad-mobile/actions/runs/30173157876) at `74c3848`:

```
[Passed] launch-smoke (10s)
1/1 Flow Passed in 10s
```

- `launch smoke (no backend)` — **success**
- `full suite (needs a seeded backend)` — **skipped**, as designed
- `test` (jest) — **success**

That is the real thing: a release APK built on the runner, installed on the emulator, launched cold while signed out, and asserted onto the login screen. Artifact `maestro-smoke-results` carries the JUnit report and the recording.

### The last fix was a build-time trap, not a test problem

The previous run failed with nothing more useful than "Sign in with Gumroad never appeared." The app was aborting on the first JS tick: `lib/env.ts` asserts `EXPO_PUBLIC_GUMROAD_CLIENT_ID` and `EXPO_PUBLIC_MOBILE_TOKEN` are defined and throws at import time if either is missing, so the process died before the first frame rendered.

The part worth remembering: **Expo inlines every `EXPO_PUBLIC_*` value into the JS bundle at build time, not at run time.** So they have to be present when Gradle builds the bundle — setting them only on the job is not enough, they also have to be on the step that runs the Gradle build. And locally the gitignored `.env` supplies them, which is exactly why a missing value is invisible until CI builds without one.

Both are placeholders, and they are safe to be placeholders: the smoke flow stops at the login screen and never begins an OAuth exchange, so no real client id or token is involved. Both env blocks say so inline, so nobody has to re-derive it.

### Where this leaves the two jobs

**smoke** is trustworthy today and needs no secrets, which is what lets it run on forks and on every PR. It catches the class of failure jest cannot see at all — a build that won't boot, a crash on launch, a broken root layout, a routing change that drops a signed-out user on a blank screen.

**suite** stays on `workflow_dispatch` and fails fast with an explanatory error until `E2E_GUMROAD_URL` exists. I checked rather than assumed: the seeded accounts the 14 flows sign in as (`mobile_buyer_do_not_edit@gumroad.com`, `mobile_seller1_do_not_edit@gumroad.com`) are absent from the live database, staging returns 503/000, and the repo has no secrets set. Pointing it at `pull_request` today would pin a permanently red check on every PR, which just teaches everyone to ignore it. What it would take is written into the workflow header.

`isDraft` false, mergeable/clean, no unresolved review findings. Ready for your read, @gianfrancopiana.

gumclaw:
@gianfrancopiana automated review-SLA nudge: this PR has been waiting 32h for review. Goal is 24h turnaround, 7 days a week (Sahil, 2026-07-21). If it's blocked on the author, comment what's needed and gumclaw will action it same-day.

Issue #288, opened 2026-07-09
Audio playback doesn't resume from saved position on iOS; failed resume then overwrites saved progress


gumclaw:
## What

A buyer reports the iOS app does not resume audiobook playback where they left off. Every time they close and reopen the app (or the book), playback starts at 0:00 and they have to scrub manually. Their audiobook is ~4 hours long, so this makes the app nearly unusable for them.

## Why

Long-form audio (audiobooks, courses) is a core mobile use case. This is the audio equivalent of the video regression fixed in #239.

## What I verified from production data

Buyer: purchase of "Dark Psychology and Manipulation: Corporate Machiavelli…" bundle (purchase confirmed successful, Apr 11 2026).

Server-side saving WORKS. The buyer has fresh `MediaLocation` rows, all `platform: iphone`:

- file 12677849 → location **1501s**, consumed_at Jul 9 02:31 UTC
- file 17676490 → location **1012s**, consumed_at Jul 9 02:04 UTC
- file 13314860 → location **1s**, consumed_at Jul 9 12:54 UTC

So the write path (`POST api/mobile/media_locations`) is recording progress. The failure is on the **resume/read side** in the app, plus a **progress-clobbering side effect**:

1. **Resume paths that never seek:** the only place the app seeks to the saved location is `playAudio({ resumeAt })` in `components/use-audio-player-sync.ts`, and `resumeAt` comes from the webview click payload (`MobileAppAudioFileRow` state in the Rails `Download/FileList.tsx`). Any playback started outside that path — lock-screen/notification controls, the mini player, or the queue restored by `initFromPlayer()` after a cold start — begins at position 0 with no seek to the saved `MediaLocation`. The webview's `latestMediaLocation` state can also be stale if the download page isn't re-fetched after reopening the app.

2. **Failed resume then clobbers the real progress:** `components/track-player-service.ts` `syncCurrentPosition()` (fires on `RemotePause` / `RemoteStop`) has no minimum-position guard, unlike `syncMediaLocation` in `use-audio-player-sync.ts` (which skips `0 < position < 3`). So when the app restarts a track at 0:00 and the user pauses from the lock screen a second later, it POSTs `location: 1` and overwrites the saved position — exactly what the buyer's data shows on file 13314860 (`location: 1` saved at 12:54 UTC, minutes after they had real progress on the other files). This turns "doesn't resume" into "loses my place permanently."

## Steps to reproduce (expected)

[...]

PR #336, opened 2026-07-31
Resume video playback from the buyer's saved position

### LINKED ISSUE #334: Video playback never resumes from saved position (progress is saved, app never reads it back) — and a 0:00 sync can clobber it
## What

Video playback in the mobile app always starts at 0:00 instead of resuming from the buyer's saved position. Progress **is** being written to the server correctly — the app just never reads it back for video.

Reported by a long-time buyer (account since Nov 2024, Android, ~24 purchases of video poker courses) on July 30: *"gumroad used to remember and play from where I last watched, now it just sends me to the beginning."* Helper ticket: https://help.gumroad.com/all?id=1e791315a05b3ab4b70284e8e4ebc57a

This is the **video** equivalent of #288 (audio resume, iOS, closed July 10). Distinct bug, distinct code path.

## Root cause

`app/purchase/[token].tsx` routes video taps to the player with:

```ts
initialPosition: message.payload.resumeAt ?? undefined,
```

`resumeAt` comes from the WebView click payload, which the Rails download page builds in
`app/javascript/components/Download/Interactions.tsx` (`NativeAppLink`). Only the **audio**
`TrackClick` call sites in `app/javascript/components/Download/FileList.tsx` pass a `resumeAt`
prop (lines ~509 and ~560). The video call sites — `eventName="stream_click"` (~line 302) and
`eventName="watch"` (~line 768) — pass no `resumeAt`, so the payload carries `resumeAt: null`
and `initialPosition` is always `undefined`. `app/video-player.tsx` then initialises
`currentPosition` to `0` and never seeks.

The audio path avoids this because `[token].tsx` also falls back to the API data
(`resumeAt: fileData.latest_media_location?.location` when building `allAudioTracks`, line ~189)
and the PDF path does the same (`initialPage: fileData?.latest_media_location?.location ?? message.payload.resumeAt`,
line ~170). The video branch has **no such fallback** — it trusts the WebView payload alone.

So there are two independent holes, either of which fixes this:

1. **Mobile (recommended, self-contained):** in the video branch of `[token].tsx`, mirror the PDF/audio
   pattern — `initialPosition: fileData?.latest_media_location?.location ?? message.payload.resumeAt`.
   `latest_media_location` is already serialised for every file by
   `UrlRedirect#mobile_product_file_json_data` (gumroad `app/models/url_redirect.rb`), so no API change
   is needed.
2. **Web (optional, also fixes the older native iOS/Android bridges):** pass `resumeAt={latestMediaLocation}`
   on the `stream_click` and `watch` `TrackClick`s in `FileList.tsx`.

Worth also confirming `video-player.tsx` guards against a saved position at/past the end
(the audio path has `isResumableLocation` in `lib/media-location.ts`; the video path has no equivalent,
so a completed video would seek to its end and appear to not play).

## Evidence (production, audited console read)

Buyer's `MediaLocation` rows are healthy and current — the write path works on both platforms:

| purchase | file | location | unit | platform | consumed_at |
|---|---|---|---|---|---|
| 348569596 | 43370573 | 1209 | seconds | android | 2026-07-30 14:51 UTC |
| 348569596 | 43370573 | 1107 | seconds | web | 2026-07-30 13:21 UTC |
| 328231798 | 16793446 | 2082 | seconds | android | 2026-07-29 16:24 UTC |
| 267616683 | 19035310 | 4551 | seconds | android | 2026-07-16 15:18 UTC |
| 328231799 | 22163347 | 3669 | seconds | android | 2026-07-16 10:02 UTC |

Every file involved is `filegroup: video`, `stream_only: true`, mp4. Matching
`ConsumptionEvent` rows are `event_type: watch, platform: android`.

Note the `loc=0` rows (e.g. purchase 346771543 / file 44893837, android, July 24; purchase
328231811 / file 15076048, android, July 7): the app opened at 0:00 and the 5-second periodic
sync in `video-player.tsx` wrote `0` straight back. That is the same **clobber** class #288 fixed
for audio — the video path still has it, because `updateMediaLocation` is called from the interval
and the unmount cleanup with no `isMeaningfulLocation` guard. So buyers don't just fail to resume,
they can lose a saved position.

## Checklist

- [ ] Video branch in `app/purchase/[token].tsx` falls back to `fileData.latest_media_location.location` (mirror the PDF/audio pattern)
- [ ] `video-player.tsx` periodic sync + unmount save apply `isMeaningfulLocation` so a fresh-at-0 player can't clobber a saved position
- [ ] `video-player.tsx` skips seeking when the saved position is at/past the end (`isResumableLocation` equivalent)
- [ ] Optionally pass `resumeAt` on the `stream_click` / `watch` `TrackClick`s in gumroad `FileList.tsx` so the legacy iOS/Android WebView bridges resume too
- [ ] Verify on Android and iOS: watch a stream-only video partway, back out, reopen from the purchase page, playback resumes

gumclaw:
## What

Video playback in the mobile app resumes from the buyer's saved position instead of always starting at 0:00, and can no longer overwrite that position with 0.

Three changes:

1. **Read the saved position back.** The video branch of `app/purchase/[token].tsx` set `initialPosition` from the WebView click payload's `resumeAt`, which the Rails download page only populates for audio rows — so it was always null and `app/video-player.tsx` initialised at 0 and never seeked. It now falls back to `fileData.latest_media_location.location` from the API, mirroring what the PDF branch (line 170) and the audio branch (line 189) already did, and forwards `content_length` alongside it. `app/post/[id].tsx` forwards `content_length` too.

2. **Stop the 0:00 clobber.** The player's two save paths — a 5-second interval and the unmount cleanup — wrote whatever position they found with no guard, so a player that opened at 0:00 because the resume failed wrote 0 straight back and destroyed the real saved position. Both now route through the shared `isMeaningfulLocation` from `lib/media-location.ts`, the same guard the audio path has used since #289, and a position at the end of the video saves as the full duration rather than a near-end timestamp.

3. **Don't seek to the end of a finished video.** A saved position at or past the end would seek there, playback would stop instantly, and it reads as a video that will not play. `content_length` catches that at mount; the loaded duration catches the files where `content_length` is not serialised. `readyToPlay` re-fires after every seek and rebuffer, so that reset is gated to run at most once per source — without the gate, a buyer rewatching a finished video gets yanked to 0:00 on every seek.

The player surface reports its position as an accessibility value, following the pattern the audio player's seek bar already uses, which is what lets the new Maestro flow assert the position across an app relaunch.

## Why

Fixes #334. A buyer with ~24 video-course purchases reported that the app stopped resuming where they left off. Production `MediaLocation` rows for that account are healthy and current on both Android and web, so the write path was working — the app simply never read it back for video. Audio and PDF were unaffected because both already had the API fallback; video was the one branch that trusted the WebView payload alone. This is the video-side twin of #288, which fixed the audio path.

The same account's rows also include several `location: 0` entries written from Android. Those are the clobber: the player opened at 0:00, the 5-second sync wrote 0, and the real position was gone. So the bug did not just fail to resume, it destroyed progress — which is why the guard is part of this PR rather than a follow-up.

`content_length` is safe to compare against a position in seconds: `ProductFile#content_length` returns `duration` for any `streamable?` file, and `filegroup == "video"` is exactly what `streamable?` means.

## Before/After

**Before:** every video started at 0:00 regardless of saved progress, and a failed resume overwrote the saved position with 0.

**After:** the video opens at the saved position; a finished video restarts from the beginning; a player that never got past the start does not write anything.

No pixels changed — this PR touches playback position logic and adds an accessibility value. There is no new or altered UI surface to photograph.

**Video walkthrough: outstanding for a maintainer.** I cannot record one honestly. Capturing this behaviour needs a real device or emulator playing a real purchased video, backgrounding the app, and reopening it; no Android emulator or booted iOS simulator is available in this environment, and a mocked capture would not demonstrate the thing under test (the position surviving a relaunch). The `.maestro/video-restart-resume.yaml` flow added here is the executable form of that same walkthrough — it plays partway, relaunches the app, and asserts the position came back — and it fails on the pre-fix code twice over, since the accessibility value did not exist and the resumed position read 0:00. Note the Maestro suite is release-QA-only today: the seeded `*_do_not_edit@gumroad.com` accounts do not exist in production and the flows open the Expo dev client, so it cannot gate CI yet.

## Test Results

Test results

70 passing across the two affected suites; full local suite 510 passing / 52 suites, `npx tsc --noEmit` clean, eslint 0 errors on touched files.

Every new test was checked red-first rather than assumed load-bearing. Reverting `app/video-player.tsx` to main reddens 5 of them; reverting `app/purchase/[token].tsx` reddens 2 of the purchase-screen tests; and mutating the re-entry guard alone (dropping `!savedPositionIsAtEnd`) reddens exactly the rewatch test, so that guard is pinned rather than incidentally green.

CI first runs here on the PR rather than on the branch: unlike antiwork/gumroad, this repo's `test.yml` and `e2e.yml` trigger only on `pull_request` and `push: branches: [main]`, so a pushed feature branch runs no workflows at all.

## QA steps

1. As `mobile_buyer_do_not_edit@gumroad.com`, open **Mobile Test Product 1** from the Library.
2. Tap **Watch** on *Mobile Test Video*, let it play a few seconds, then go back.
3. Kill and reopen the app, return to the product, and tap **Watch again** — playback should pick up where it stopped, not at 0:00.
4. Let a video play all the way to the end, leave, and reopen it — it should restart from the beginning and play, not sit frozen at the end.
5. Open a video, immediately go back without letting it play, then reopen — the previously saved position should still be there rather than reset to 0:00.
6. Or run it: `maestro test -p android -e APP_ID=$ANDROID_BUNDLE_NAME .maestro/video-restart-resume.yaml`

## Review

A local fable-5 adversarial review ran against the branch before this PR was opened and returned SHIP-WITH-FIXES with one P1 and three P3s, all fixed in `5068f11`:

- **P1** — the finished-video reset re-fired on every `readyToPlay`, and that event is not one-shot (Media3 re-enters it after each seek and rebuffer; iOS re-derives it from buffer state), so a buyer rewatching a finished video was snapped back to 0:00 on every seek. This was a regression the first commit introduced. Fixed by gating the reset on state, with a regression test proven by mutation.
- **P3** — the at-end state was not recomputed when the media changes in place, suppressing a legitimate resume on the next video.
- **P3** — a 0 or NaN duration at `readyToPlay` overwrote a known `content_length` and rendered "NaN:NaN".
- **P3** — the Maestro flow assumed a starting position, but its own previous runs and the sibling `video-playback.yaml` move that position forward on a shared seeded account, so it would have rotted after a few QA runs. It now normalises by playing to the end first. The seeded fixture is 10 seconds (`Big Buck Bunny - Trailer.mp4`), which is why every asserted position sits between 3 seconds and the end.

The review also independently confirmed that `updateMediaLocation` has exactly one call site in the player reached from exactly two paths, both now guarded, and that the `"0"`-string cases through expo-router params are handled.

Greptile then found a media-switch defect the fable review missed, fixed in `55216dc`: switching to another video left the previous video's position and duration in component state, so the screen briefly showed the old position and — while the new video's stream lookup was still in flight — a periodic save could write that old position against the new video's file, making the replacement video reopen at the wrong place. Per-video progress state now resets on a media change, and both save paths bail out until the player actually holds the media the screen's parameters describe. Two tests cover it, each proven by mutation: dropping the state reset reddens the position-display test, and dropping the identity guard reddens the save test with the exact bug shape (`location: 600` from the previous video written against `productFileId: file-2`).

---

## AI disclosure

Written by **claude-opus-5** (Anthropic) running as a Hermes agent, from this instruction on issue #334:

> Fix in PR and have maestro test for this

The agent read the issue's diagnosis, confirmed both holes in the code, verified in the Rails repo that `content_length` is a duration in seconds for video files, implemented the fix, wrote the tests and verified each one red-first, authored the Maestro flow, ran a local fable-5 adversarial review on the branch, and fixed the P1 and P3s that review found. The video walkthrough is genuinely outstanding, as described above.


gumclaw:
Fixed in `55216dc` — the media-switch finding was real.

Two holes, both now closed. `currentPosition`/`videoDuration` and the playback-started flag survived a media change, so the screen showed the previous video's position; and the 5-second save plus the unmount save read straight off the player, which still holds the previous source while the next video's stream lookup is in flight, so a save landing in that window wrote the old position against the new video's `product_file_id`. Progress state now resets in the media-change branch, and both save paths bail out until the player's resolved media matches the screen's current parameters.

Two tests cover it and each was checked by mutation rather than assumed load-bearing. Dropping the state reset reddens the position-display test; dropping the identity guard reddens the save test with exactly the shape you described — `location: 600` from the previous video written against `productFileId: file-2`.

Suite is 70 passing on the two affected files, 510 passing across 52 suites locally, `tsc --noEmit` clean, eslint 0 errors on touched files. PR body updated with the same detail.

gumclaw:
@nyomanjyotisa gentle nudge — this has been on your review queue for 48 hours now (requested 2026-07-31T13:46Z) and nothing is outstanding on my side.

State at head `55216dc`: all 3 checks green, MERGEABLE, no unresolved threads. Greptile's one real finding (saved position not being reset when the player switches media) was fixed in `55216dc` and written up in my 07-31T16:08Z comment.

What's needed: a review pass, and the merge press if it looks right. If it's capacity rather than doubt, say the word and I'll re-route to @gianfrancopiana, who has been merging in this repo lately. This is my first and only ping on it for the next couple of days.

PR #338, opened 2026-08-02
Treat an HTML 404 as a lost session so Android can refresh its token

gumclaw:
## What

Treats an **HTML 404** as a lost session, so the token-refresh path runs instead of a misleading `404 Not found` reaching the screen.

Everything this app requests answers in JSON. An HTML 404 therefore means the request never reached the endpoint it asked for — it was redirected to a web page that does not exist on that host. One new predicate in `lib/request.ts` reads `content-type` off the response:

```ts
const isHtmlResponse = (response: Response): boolean =>
  response.headers.get("content-type")?.toLowerCase().includes("text/html") ?? false;
```

and the unauthorized branch gains `(response.status === 404 && isHtmlResponse(response))`.

## Why

`GUMROAD-MOBILE-Z5` is 548 events across **458 users in 14 days**, still firing at ~53/day, and **every current-build event is Android**. antiwork/gumroad-mobile#335 fixed the redirect detection for iOS, but that detection compares the request URL against the final URL — and the final URL is structurally unavailable on Android:

- **iOS** — `RCTNetworking.mm:621` emits `response.URL.absoluteString`, the URL the redirect landed on.
- **Android** — `NetworkingModule.kt:635-639` passes the closure-captured `url` parameter of `sendRequestInternalReal`, i.e. the **originally requested** URL. `NetworkEventUtil.onResponseReceived` even names the parameter `requestUrl`.

So on Android `response.url` always equals the requested URL, and any redirect-URL-based check compares a path against itself. A different signal was needed, and the content type is one the platform cannot take away.

**Why content-type is the right signal, measured against production rather than assumed:**

| Request | Status | Content-Type | Bytes |
|---|---|---|---|
| `GET https://api.gumroad.com/mobile/nonexistent.json` (genuine missing endpoint) | 404 | `application/json` | 34 |
| `GET https://api.gumroad.com/mobile/analytics/products.json` (unauthenticated) | 401 | `application/json` | 45 |
| `GET https://api.gumroad.com/login` (where an unauthenticated read is redirected) | 404 | **`text/html`** | 900 |
| `GET https://gumroad.com/oauth/token` (a refresh POST downgraded to GET) | 404 | **`text/html`** | 900 |
| `POST https://gumroad.com/oauth/token` (correct refresh) | 400 `invalid_grant` | `application/json` | 213 |

The two rows that lose a session are exactly the two HTML rows, and a genuine 404 from the API is JSON. The split is clean, so the new branch cannot swallow a real missing-endpoint error.

This also covers the majority mechanism, which the redirect check never could. `auth_path=refresh_failed` is set on **898 of 1631** events, and that tag is attached only when capturing the refresh error itself — so those 404s came from the refresh call to `/oauth/token`, not from a redirected API GET. The Sentry breadcrumbs show `method: GET` with `status_code: 404` and `response_body_size: 1253` on a call the code makes as a POST; `GET /oauth/token` is precisely a 900-byte HTML 404, and Cloudflare's RUM beacon adds the 353-byte delta when the client sends a browser-shaped `Accept`. Both the tag and the method fit.

### Alternatives rejected

- **Match on the response body** (`<html` / "The page you were looking for") — brittle against copy changes and needs the body read before the status branch, which would break the purged-blob handling in `readBody`.
- **Special-case the `/oauth/token` path** — fixes the refresh leg only, leaves every redirected API read broken on Android, and hardcodes an endpoint.
- **Follow redirects manually with `redirect: "manual"`** — whatwg-fetch does not support it; XHR follows redirects in the native layer with no JS hook on either platform.
- **A native module reading `response.request().url()` on Android** — correct but far larger, needs a new build, and reaches no user until the next store release.

## Before / After

| | Before | After |
|---|---|---|
| Android API read whose session expired | `RequestError: Request failed: 404 Not found`, screen fails, session never refreshed | `UnauthorizedError` → `useAPIRequest` refreshes the token and retries the read |
| Android token refresh answered by the HTML 404 | Same generic 404, reported to Sentry as `auth_path=refresh_failed` | `UnauthorizedError` → `console.warn` + clean logout, no Sentry noise |
| Genuine missing API endpoint (JSON 404) | `Request failed: 404 Not found` | unchanged — `Request failed: 404 Not found` |
| iOS redirected read | already handled by #335 | unchanged |

There is no pixel to photograph: this changes **which error class is thrown**, not what renders. The evidence is the mutation proof below plus the production content-type table above.

## Walkthrough

https://github.com/antiwork/gumroad-mobile/blob/fdece264f8090a5f30219afa6c61922fbc228b3f/qa-media/pr-338-html-404-session-walkthrough.mp4?raw=true

Five sections, all against real values rather than a spec summary: the live production probes producing the content-type table above, the diff, the 20-green run, the mutation proof reverting `isHtmlResponse` to `false` and reddening exactly the two new examples, and the restore back to 20 green.

mutation proof

## Test Results

Two new examples, both mutation-verified against a neutered predicate rather than merely passing:

```
$ npx jest tests/lib/request.test.ts
Tests: 20 passed, 20 total

# fix reverted (isHtmlResponse forced to return false):
  ✕ throws UnauthorizedError on an HTML 404 even when the response was not redirected
  ✕ throws UnauthorizedError when a token refresh POST comes back as an HTML 404
Tests: 2 failed, 18 passed, 20 total
```

Exactly the two new examples redden, and the pre-existing 18 stay green — including `does not treat a direct JSON 404 from /login as unauthorized` and `still throws the generic 404 error for a JSON 404 from the API`, which are the guards against this branch being too greedy.

Also run: `npx prettier --check` (clean after formatting the wrapped condition), `npx tsc --noEmit` (one pre-existing `@rn-primitives/portal` resolution error in `app/_layout.tsx`, unrelated and present on `main`).

The existing mocks in `tests/lib/request.test.ts` had no `headers`, so each was given a minimal `headers.get` stub carrying the content type that endpoint really returns in production — `text/html` for the two `/login` redirect cases, `application/json` elsewhere.

## QA steps

1. On an **Android** build, sign in and let the access token expire (or revoke it server-side).
2. Pull to refresh Library or Analytics.
3. Expect the session to refresh and the data to load, instead of a generic failure.
4. With the refresh token also revoked, expect a clean logout to the login screen rather than a `404 Not found` error and a stuck session.
5. Confirm a genuine missing endpoint still surfaces its own error — no behaviour change there.
6. Re-check `GUMROAD-MOBILE-Z5` after the release ships; this is the fix that should move its numbers.

## Post-release watch point

If `/oauth/token` is ever **misrouted by a deploy** — a live endpoint answering with an HTML 404 rather than an expired session — every user hitting it is logged out. Worth naming, though it is not a regression: `useAPIRequest` and `useAuthedRequest` already call `logout()` unconditionally on any refresh failure that is not `KeychainUnavailableError` (`lib/request.ts:250-261`), so that outcome is today's behaviour too. This PR changes only which arm runs — `console.warn` instead of `Sentry.captureException`. The mitigation is alerting on the endpoint, not narrowing the predicate, which would reintroduce the Android blind spot.

## Not in this PR

The remaining open question on antiwork/gumroad-private#1616 is *why* the refresh POST arrives as a GET at all. The `www` → apex 301 was ruled out (following it lands on an 83-byte JSON 400, not the 1253-byte HTML 404). This PR makes the app recover correctly from that state regardless of its origin; identifying the redirector is still owed and stays on the issue.

---

**AI disclosure:** Written by claude-opus-5 running as Hermes Agent (Gumclaw). Instructions given: act as the autonomous dev dispatcher for antiwork/gumroad-private, sweep for human replies on open issues, and action them — @slavingia replied "Build" on gumroad-private#1616. The agent read the React Native `NetworkingModule.kt` and `RCTNetworking.mm` sources to confirm the Android `responseURL` limitation, probed the production API and web hosts with curl to establish the content-type split above, wrote the two new tests, and verified them by reverting the predicate and confirming exactly those two reddened.


gumclaw:
Self-review, having gone looking for the ways this could be wrong rather than for reasons it is right.

**Could the new branch swallow a real error?** It fires only on `404` **and** `text/html`. Every consumer of `request()` reaches it through `requestAPI`/`useAPIRequest`/`useAuthedRequest`, and the API answers a genuinely missing endpoint with `application/json` (measured: 34 bytes, in the video). Two existing examples pin that — `does not treat a direct JSON 404 from /login as unauthorized` and `still throws the generic 404 error for a JSON 404 from the API` — and both stay green under the mutation. The only place a 404 is load-bearing as a *value* is `use-agent-meta.ts`, and that keys on 403, not 404.

**Does an `UnauthorizedError` here do something harmful?** It routes into `useAPIRequest`'s catch and `useAuthedRequest`, both of which attempt one refresh and then log out. Today those same requests throw `RequestError` and fail the screen with the session still nominally alive, which is strictly worse: the user is stuck rather than re-authenticated. The one behaviour change worth naming is that a *persistent* HTML 404 now ends in a logout instead of an error screen. That is correct if the cause really is a lost session, and this PR's own evidence is what argues it is — but it is the thing to watch after release, so QA step 6 says to re-check `GUMROAD-MOBILE-Z5` rather than assume.

**Is the test actually load-bearing?** Verified by neutering `isHtmlResponse` to `false`, not by assuming: exactly the two new examples redden, 18 stay green. If the predicate had been over-broad, the JSON-404 guards would have failed instead.

**Weakest point, stated plainly:** this is a *recovery* fix, not a root-cause fix. Why the refresh POST arrives as a GET is still open on antiwork/gumroad-private#1616 (the `www` 301 is ruled out — following it lands on an 83-byte JSON 400, not the HTML 404). If the real cause turns out to be something that also breaks the refresh itself, this makes the failure clean rather than absent.

@claude review once

gumclaw:
Thanks — I checked this one against the caller code rather than the description, and the premise does not hold. Not merging over it, but I want the reasoning on the record rather than a bare disagreement.

**The claim is that an HTML 404 from the OAuth token endpoint "can trigger refresh/logout behavior rather than preserve the not-found failure."** The refresh path already logs out on *every* failure that is not `KeychainUnavailableError` — `lib/request.ts:250-261`:

```ts
} catch (refreshError) {
  if (refreshError instanceof KeychainUnavailableError) throw error;
  if (refreshError instanceof UnauthorizedError || refreshError instanceof SessionExpiredError) {
    console.warn(refreshError);
  } else {
    Sentry.captureException(refreshError, { tags: { auth_path: "refresh_failed" } });
  }
  await logout();      // <- unconditional, both before and after this PR
  throw error;
}
```

`lib/authed-request.ts:16-26` is the same shape. So a `RequestError` and an `UnauthorizedError` from that call produce **identical** logout behaviour. The only thing this PR changes on that leg is which arm of the `if` runs: `console.warn` instead of `Sentry.captureException`. That is the point — those are the **898 of 1631** events currently tagged `auth_path=refresh_failed`, and they are session losses being reported as crashes.

**The blast radius is fully enumerable, and it is three call sites.** `request()` is only reached directly from `lib/auth-context.tsx:48` (products endpoint, API host, answers JSON), `:135` (authorization-code exchange) and `:216` (refresh). Everything else goes through `requestAPI` to the API host. On `:135` the delta is also only the error class in a `Sentry.captureException` — no logout there either, since the user has no session yet.

**On narrowing to known hosts/paths, which is the suggested fix:** that reintroduces the bug this PR exists to remove. The Android population cannot be identified by route — `NetworkingModule.kt:635-639` reports the *requested* URL as `responseURL`, so there is no redirect destination to validate, which is exactly why antiwork/gumroad-mobile#335 could not reach these users. Hardcoding the two known routes also means the next surface that starts redirecting is broken again silently. The content-type split is the general form of the same rule, and it is measured, not assumed: a genuine missing API endpoint answers `application/json` (34 bytes), the two session-loss routes answer `text/html` (900 bytes). Both directions are pinned by tests that stay green under the mutation.

**Where I think the finding is pointing at something real:** if `/oauth/token` were *misrouted by a deploy* — a live endpoint returning an HTML 404 rather than an expired session — every affected user gets logged out. That is worth naming, and I have added it to the PR body as the post-release watch point. But it is not a regression: that outcome is what the unconditional `logout()` above already produces today. It argues for alerting on the endpoint, not for narrowing the predicate.

gumclaw:
Draft-completion check — readied.

All three checks are green at `fdece264f` and there is no stated blocker left. The body already carries everything the gates ask for: Test Results with a mutation proof (20 green; reverting `isHtmlResponse` to `false` reddens exactly the two new examples and nothing else), a production content-type probe table showing the two session-losing responses are exactly the two `text/html` rows while a genuine missing endpoint answers JSON, a walkthrough video and the mutation-proof image committed at head, a rejected-alternatives section, and the AI disclosure.

No hosted preview is possible in this repo (Expo — the deployments API returns zero for the whole repo), and there is no pixel to photograph regardless: the change alters which error class is thrown, not what is drawn.

Assigned to @nyomanjyotisa, no review request yet.

gumclaw:
Momentum sweep — green and evidenced, so this is going to final human review now.

3/3 checks green at head, MERGEABLE, and the QA evidence is committed in `qa-media/` rather than described. Per the convention that a review request is the last step and not part of opening a PR, that bar is met, so I have requested review and set the `awaiting-human` label to say the ball is on your side.

Nothing outstanding from me; I am not arming auto-merge.

PR #339, opened 2026-08-02
Send proposal ids from mobile Agent confirms

gumclaw:
## Status

- [x] Scope — gp#1727 identifies mobile confirms dropping `proposal_message_id`, forcing the server onto legacy fuzzy matching.
- [x] Design — carry the server's top-level `proposal_message_id` onto the mobile `ProposedAction`, then include it in the confirm payload alongside `conversation_id`.
- [x] Build — branch `gp1727-mobile-proposal-id`, head `a6a193b`.
- [ ] Ship — draft PR; CI is running and the mandatory PR video is still pending/not applicable for a non-visual data-layer fix.
- [x] Market — n/a; bug fix for mobile Agent confirm binding.
- [x] Sell — n/a; no pricing or sales collateral change.

## What

Mobile Agent confirms now send `proposal_message_id` back to `/mobile/agent/actions` when the stream provided one. The done-frame parser attaches the id to the proposed action, and the confirm mutation posts it back with the existing action payload.

## Why

Without the proposal id, every mobile confirm falls back to the server's legacy payload matcher. When a seller stages two near-identical product proposals, two separate “confirm” taps can match two different pending rows and create duplicate products. Web already sends this id; mobile should use the same binding.

Issue: antiwork/gumroad-private#1727

## Before/After

Before:
- The SSE `done` frame's `proposal_message_id` was ignored.
- `executeAgentAction` sent only `type`, `params`, and optional `conversation_id`.
- The server logged `Store agent confirmation omitted proposal message id` and used fuzzy matching.

After:
- `streamAgentMessage` preserves `proposal_message_id` on the proposed action.
- `executeAgentAction` includes `proposal_message_id` in the confirm POST.
- Existing confirms without an id still work through the compatibility path during deploy skew.

## Test Results

- `npm exec -- jest tests/lib/agent.test.ts tests/components/agent/agent-chat.test.tsx --silent --forceExit` — passed, 44 tests.
- `npm exec -- tsc --noEmit --pretty false` — passed.
- `npm exec -- eslint lib/agent.ts tests/lib/agent.test.ts tests/components/agent/agent-chat.test.tsx` — passed.
- `npm exec -- prettier --check lib/agent.ts tests/lib/agent.test.ts tests/components/agent/agent-chat.test.tsx` — passed.
- `claude-fable-5` adversarial review — clean for P1/P2. It noted two non-blocking test-hygiene points; both were fixed in `a6a193b`.

## QA steps

No UI change. The covered path is: mobile Agent stream receives a proposed action with `proposal_message_id`, the seller taps Confirm, and the confirm request sends that exact id back to the API. Verified through `tests/lib/agent.test.ts` and `tests/components/agent/agent-chat.test.tsx`.

Video: not included yet; this is a non-visual data-layer fix. If the repo requires a video for every PR despite no UI delta, the recording can show the Agent confirm flow once a dev app is running.

---

AI assistance: Gumclaw used claude-opus-5 to inspect the issue, edit the code, and update tests.

### DIFF
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -17,6 +17,7 @@ export interface ProposedAction {
   type: "api_write";
   params: Record<string, unknown>;
   summary: string;
+  proposal_message_id?: string;
   title?: string;
   fields?: ProposedActionField[];
 }
@@ -73,6 +74,7 @@ export const executeAgentAction = async ({
     data: {
       type: action.type,
       params: action.params,
+      ...(action.proposal_message_id ? { proposal_message_id: action.proposal_message_id } : {}),
       ...(conversationId ? { conversation_id: conversationId } : {}),
     },
     accessToken,
@@ -95,6 +97,7 @@ export interface AgentStreamResult {
 interface DoneEventData {
   reply: string;
   proposed_action: ProposedAction | null;
+  proposal_message_id?: string;
   conversation_id?: string;
 }
 
@@ -158,7 +161,12 @@ export const streamAgentMessage = async ({
           const data = raw as DoneEventData;
           return {
             reply: data.reply,
-            proposedAction: data.proposed_action,
+            proposedAction: data.proposed_action
+              ? {
+                  ...data.proposed_action,
+                  ...(data.proposal_message_id ? { proposal_message_id: data.proposal_message_id } : {}),
+                }
+              : null,
             conversationId: data.conversation_id ?? conversationId ?? null,
           };
         }
--- a/tests/components/agent/agent-chat.test.tsx
+++ b/tests/components/agent/agent-chat.test.tsx
@@ -112,6 +112,7 @@ describe("AgentChat", () => {
         type: "api_write",
         params: { code: "LAUNCH", percent_off: 20 },
         summary: "Create a 20% off code called LAUNCH",
+        proposal_message_id: "msg-123",
       },
     });
     mockExecuteAgentAction.mockResolvedValue("Created discount LAUNCH.");
@@ -137,6 +138,7 @@ describe("AgentChat", () => {
         type: "api_write",
         params: { code: "LAUNCH", percent_off: 20 },
         summary: "Create a 20% off code called LAUNCH",
+        proposal_message_id: "msg-123",
       },
     });
   });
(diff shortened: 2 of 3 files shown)


gumclaw:
Momentum sweep — **draft-completion check: readied.**

The only thing holding this in draft was the Ship line saying the mandatory video was "pending/not applicable", which is not a blocker so much as an unresolved question. Resolving it: **not applicable**, and stated as such rather than left ambiguous. This changes one field in a JSON request body — there is no pixel delta to record, and a screen recording of the Agent confirm flow would look identical before and after. The behaviour that matters is only visible in the request payload, which is what the tests assert.

Everything else was already in place:

- All 3 checks green at `a6a193b` — `test`, `launch smoke (no backend)`, `Greptile Review`.
- Greptile: **5/5, no issues found.** It exercised both paths, confirming `msg-123` reaches the confirm POST unchanged and that identifier-absent confirms keep the legacy payload — which is the deploy-skew case this needs to not break.
- Test Results, before/after, and AI disclosure all present. `claude-fable-5` clean for P1/P2.

`MERGEABLE`. @gianfrancopiana stays assignee. No `working` label exists in this repo, so there is nothing to flip.

gumclaw:
Closing this as superseded by #340, which is the same gp#1727 fix opened about three hours later and is a strict superset.

Both make mobile confirms send `proposal_message_id` so the server stops falling back to the legacy fuzzy payload matcher. The difference:

- **This PR** carries the id inside the `ProposedAction` object and only touches the SSE `done`-frame parser in `lib/agent.ts`. A proposal that arrives through hydrated history (`agent_message_props`) or through the non-streaming `Api::Mobile::AgentController#create` response still confirms without an id, because those paths never go through the stream parser.
- **#340** threads the id as a separate top-level field through all three server sources — the `done` frame, the non-streaming create response, and hydrated history — and wires it through `agent-chat.tsx` to the confirm mutation, so the binding holds no matter how the card got on screen. It also ships a committed QA harness (`qa-media/pr-mobile-proposal-message-id-walkthrough.sh`) that runs the real chat component against `origin/main` and against the branch and prints the actual POST body, so the before/after is a read rather than a claim.

Nothing is lost by closing this: #340 covers everything here. Branch `gp1727-mobile-proposal-id` is not deleted, so this is one click to reopen if I've misjudged the overlap.

Two sweeps opened the same fix a few hours apart — that's on me, and I'm fingerprinting new PRs against their tracker issue before treating them as independent.

PR #340, opened 2026-08-02
Send proposal_message_id on agent confirm so a double tap stops creating two products

gumclaw:
## What this fixes

`gumroad-private#1727`. The mobile client never sent `proposal_message_id` back when confirming a staged action, so **every** mobile confirm fell through to the server's legacy fuzzy-match path in `claim_agent_action` — 1,144 `Store agent confirmation omitted proposal message id` notices in 5 retained log days, against 1,812 `/mobile/agent/actions` calls over the same window.

That path can only tell proposals apart by comparing payloads. When the model stages the same intent twice, each confirm matches its own stored proposal, both claims succeed, and both replay against the v2 API. On 2026-08-02 that gave one seller two identical $25 products from two taps (Links 13693563 and 13693570; the duplicate was soft-deleted for her).

The identity binding is what makes confirm idempotent. Without it mobile gets none of the guarantees the web client has — including `ACTION_ALREADY_CONFIRMED`, which cannot fire because the second confirm resolves to a *different* row that is still pending.

## What changed

The server already emits `proposal_message_id` in three places — the SSE `done` frame (`Api::Mobile::AgentStreamsController#create`), the non-streaming create response (`Api::Mobile::AgentController#create`), and hydrated history (`agent_message_props`). **Only the client was dropping it.** No server change is needed; `claim_agent_action` already prefers the id when present.

- `lib/agent.ts` — parse `proposal_message_id` off the `done` frame into `AgentStreamResult`, accept it on `ConversationMessage` for resumed history, and send it in `executeAgentAction`'s payload.
- `components/agent/agent-chat.tsx` — carry it on `DisplayMessage` and pass it through `confirmAction`.

Deliberately **not** in this PR: the `client_turn_id` / turn-status recovery path. The mobile endpoint already exists (`Api::Mobile::AgentController#turn_status`) but wiring the client to it is a separate, larger change, and the issue lists it as a follow-up.

## Verification

Full suite, not just the touched files:

```
Test Suites: 52 passed, 52 total
Tests:       493 passed, 493 total
```

`tsc --noEmit` clean. `expo lint` reports 0 errors (one pre-existing unused-var warning in `app/login.tsx`, untouched here). Prettier clean on all changed files.

**The new test genuinely fails without the fix** — I checked rather than assuming. Applying only the test changes on top of unmodified source:

```
✕ sends the proposal message id back when confirming a streamed proposal
  Expected: ObjectContaining {"conversationId": "conv-123", "proposalMessageId": "msg-abc"}
  Received: {"accessToken": "test-token", "action": {...}, "conversationId": "conv-123"}
```

Worth saying: I originally wrote a *second* test covering a proposal restored from stored history, and it passed on unmodified `main` too. A restored proposal renders as dismissed, so Confirm never appears and the test could not fail. I deleted it rather than ship a test that guards nothing.

## QA media

[`qa-media/pr-mobile-proposal-message-id-walkthrough.txt`](qa-media/pr-mobile-proposal-message-id-walkthrough.txt), generated by the committed script beside it. One harness, run against `origin/main` and against this branch. It renders the real chat component, streams a proposal whose `done` frame carries `proposal_message_id=msg-abc`, presses Confirm, and prints the request body the client actually built — only the network boundary is faked, so `executeAgentAction` itself runs for real. The harness never asserts on the id, so a branch that does not fix the bug reports it.

```
=== origin/main (before) ===
  server sent proposal_message_id on done frame : msg-abc
  POST body keys                                : conversation_id, params, type
  POST body proposal_message_id                 : ABSENT
  server confirm path                           : legacy fuzzy match
  same intent staged twice, two taps            : BOTH EXECUTE -> duplicate product

=== fix/mobile-agent-proposal-message-id (after) ===
  server sent proposal_message_id on done frame : msg-abc
  POST body keys                                : conversation_id, params, proposal_message_id, type
  POST body proposal_message_id                 : msg-abc
  server confirm path                           : id match (idempotent)
  same intent staged twice, two taps            : second rejected, ACTION_ALREADY_CONFIRMED
```

No device video: this is a network-payload change with no rendered surface — the UI is byte-identical before and after, and a screen recording would show two indistinguishable confirm taps. The transcript above shows the thing that actually differs. Happy to record one if you'd rather have it.

## Close when

- [ ] shipped in an app build
- [ ] `Store agent confirmation omitted proposal message id` daily count drops to ~0 for mobile while `/mobile/agent/actions` volume holds — that is the verification signal, and it can only be read after release

### DIFF
--- a/components/agent/agent-chat.tsx
+++ b/components/agent/agent-chat.tsx
@@ -24,6 +24,10 @@ const isNearBottom = ({ contentOffset, contentSize, layoutMeasurement }: NativeS
 
 interface DisplayMessage extends ChatMessage {
   proposedAction?: ProposedAction;
+  // Binds a confirm to the exact stored proposal the seller saw. Without it the server falls back
+  // to matching on payload, which cannot tell two near-identical proposals apart and has
+  // double-created products when the model staged the same intent twice.
+  proposalMessageId?: string;
   actionStatus?: "applied" | "dismissed";
 }
 
@@ -171,6 +175,7 @@ export const AgentChat = ({ greeting, suggestions }: Props) => {
               role: message.role,
               content: message.content,
               ...(message.proposed_action ? { proposedAction: message.proposed_action } : {}),
+              ...(message.proposal_message_id ? { proposalMessageId: message.proposal_message_id } : {}),
               ...(message.action_status
                 ? { actionStatus: message.action_status }
                 : // A proposal that was never confirmed is stale after resuming, so show it as dismissed.
@@ -202,11 +207,16 @@ export const AgentChat = ({ greeting, suggestions }: Props) => {
           },
         }),
       ),
-    onSuccess: ({ reply, proposedAction, conversationId }) => {
+    onSuccess: ({ reply, proposedAction, proposalMessageId, conversationId }) => {
       conversationIdRef.current = conversationId;
       setMessages((prev) => [
         ...prev,
-        { role: "assistant", content: reply, ...(proposedAction ? { proposedAction } : {}) },
+        {
+          role: "assistant",
+          content: reply,
+          ...(proposedAction ? { proposedAction } : {}),
+          ...(proposalMessageId ? { proposalMessageId } : {}),
+        },
       ]);
     },
     onError: () => {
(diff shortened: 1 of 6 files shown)


gumclaw:
Momentum sweep — green and evidenced, so this is going to final human review now.

3/3 checks green at head, MERGEABLE, and the QA evidence is committed in `qa-media/` rather than described. Per the convention that a review request is the last step and not part of opening a PR, that bar is met, so I have requested review and set the `awaiting-human` label to say the ball is on your side.

Nothing outstanding from me; I am not arming auto-merge.

PR #343, opened 2026-08-03
Recover an interrupted agent turn instead of telling the seller it failed

gumclaw:
Ref antiwork/gumroad-private#1727

## What

The mobile Agent chat now recovers a turn when its response stream breaks instead of treating the turn as failed.

Each send gets a stable `client_turn_id`. If the request or stream ends before the terminal event, the app asks the existing turn-status endpoint what happened to that exact turn. A persisted reply replaces the interrupted response, including its proposal and `proposal_message_id`; a failed or lost turn still shows the existing error.

Recovery has a hard three-minute wall-clock deadline. The deadline covers poll delays and in-flight status requests, so a slow request cannot keep the composer disabled past the budget. Leaving the screen cancels the stream, polling, and pending delays.

A proxy-driven Maestro flow now cuts the stream after its first token and verifies that the app adopts the persisted proposal without confirming it.

## Why

Before this change, a connection drop could show “Sorry, I ran into a problem” even after the server had accepted and persisted the turn. Retrying could then start a second turn for work the server had already completed.

The stream cannot resume, and sending the prompt again would create a new turn. Polling by `client_turn_id` is the only safe way to learn whether the original turn persisted without risking a duplicate proposal.

## Before/After

https://github.com/user-attachments/assets/da6ff2e1-5fa3-4e0c-88a1-381336df100e

Before: a mid-stream disconnect discards the partial response and ends with the generic failure message.

After: the partial response clears, the app shows “Working on it…” while it checks the original turn, then renders the recovered reply and proposal when the server reports `persisted`.

The recording uses an iPhone 17 Simulator on iOS 26.4. The app, authentication, and ordinary API requests point to production. The Agent stream and turn-status responses use a deterministic proxy fixture because the available production accounts receive `403` from the Agent endpoints. The fixture sends one token, cuts the socket, reports `in_progress`, then returns the persisted proposal. It validates the client recovery flow, not production Agent access or model output.

## Test Results

- Full Jest suite: 53 suites and 544 tests passed.
- Focused Agent suites: 72 tests passed.
- TypeScript: passed.
- Lint: passed with two pre-existing warnings outside this change.
- iOS Simulator recovery flow: passed; the fallback message stayed hidden and the recovered proposal rendered with Confirm enabled.
- CI `test`, `launch smoke (no backend)`, and Greptile Review: passed at `7bd5f16`.
- Claude Fable 5 autoreview with high reasoning: no findings.

---

The initial implementation used Hermes Agent with Claude Opus 5. Follow-up fixes, tests, and Simulator validation used Codex with GPT-5.6 Sol. Claude Fable 5 performed the final autoreview.

### DIFF
--- a/.maestro/agent-turn-recovery.yaml
+++ b/.maestro/agent-turn-recovery.yaml
@@ -1,0 +1,35 @@
+appId: ${APP_ID}
+tags:
+  - ios-only
+  - proxy-only
+---
+- runFlow:
+    file: utils/sign-in.yaml
+    env:
+      EMAIL: mobile_seller1_do_not_edit@gumroad.com
+      PASSWORD: password
+
+- tapOn: ".*Agent.*"
+- extendedWaitUntil:
+    visible:
+      id: "agent-message-input-ready"
+    timeout: 15000
+- evalScript: ${output.recoveryProduct = "RECOVERY_E2E_" + Date.now()}
+- tapOn:
+    id: "agent-message-input-ready"
+- inputText: ${"Create a product proposal named " + output.recoveryProduct + " priced at 1 dollar. Do not ask follow-up questions."}
+- tapOn:
+    text: "Send"
+    enabled: true
+    waitToSettleTimeoutMs: 100
+- extendedWaitUntil:
+    visible:
+      id: "agent-message-input-sending"
+    timeout: 5000
+- extendedWaitUntil:
+    visible:
+      id: "agent-message-input-ready"
+    timeout: 240000
+- assertNotVisible: "Sorry, I ran into a problem. Please try again."
+- assertVisible: ".*${output.recoveryProduct}.*"
+- assertVisible: "Confirm"
--- a/.maestro/utils/login-with-password.yaml
+++ b/.maestro/utils/login-with-password.yaml
@@ -4,40 +4,22 @@ appId: ${APP_ID}
     element: Login
 - tapOn: Email
 - inputText: ${EMAIL}
-- pressKey: enter # The form moves focus to the password field.
-# The password field rerenders after each character in Safari on iOS 26,
-# so refocus it before entering each character of the seeded test password.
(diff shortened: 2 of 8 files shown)


gumclaw:
@claude review once

gumclaw:
Fixed the recovery budget P1 in `d3733c8` — it is a wall-clock deadline now, so a hanging status request cannot stretch the window past 3 minutes. 51 tests pass and the mutation proof (reverting the recovery to a bare rethrow) reddens 2 of the 3 recovery examples; details are in the description. Still draft: the walkthrough video is owed.

gumclaw:
@slavingia you took this out of draft at 05:56Z — leaving it that way, your call stands. One note so the ready state is read accurately.

The **automated** evidence here is strong and complete: 51 jest tests pass, tsc and eslint clean, CI 3/3 green at `d3733c8`, and there's a real mutation proof in the body (replacing the recovery call with a bare rethrow kills exactly the 2 examples that should die, and the 3rd is correctly explained as guarding the error path instead).

What is **not** done is the on-device leg, listed in the body under "QA steps — Owed, on a device build": cut the network mid-stream, restore it inside the ~3-minute recovery budget, confirm the reply appears with its proposal card intact rather than "Sorry, I ran into a problem". That needs a physical build — this repo has zero deployments ever (Expo), so there is no hosted preview I can screenshot, and no amount of CI here can substitute for it.

@gianfrancopiana — that's the one thing left, and it's a few minutes on a device you already have set up. If you'd rather not, say so and I'll ask Sahil to route it. Marking `awaiting-human` since the remaining work is genuinely not mine to do.

slavingia:
Can you QA this yourself using simulator?

gianfrancopiana:
> Can you QA this yourself using simulator?

I'll be pushing improvements + E2E test shortly. But agreed @gumclaw should be able to run the simulator.

PR #344, opened 2026-08-04
Fix walkthrough script destroying uncommitted edits on interruption

gumclaw:
## What

Fixes a Greptile P1 flagged on the merged antiwork/gumroad-mobile#340: the QA
walkthrough script `qa-media/pr-mobile-proposal-message-id-walkthrough.sh`
checked out `origin/main` over `lib/agent.ts` and
`components/agent/agent-chat.tsx` to capture the "before" state, then
restored `HEAD` — permanently discarding uncommitted edits in those two files
if the run was interrupted between the two checkouts.

## Why

Anyone running this walkthrough with local, uncommitted work in those files
could lose it with no warning, and an interrupted run leaves both files
sitting at the base revision rather than restored.

## Before / after

The "before" capture now happens in a disposable, detached `git worktree`
checked out at `origin/main` (with `node_modules` borrowed via symlink); the
caller's checkout is never touched. The "after" run still executes in place
against this branch's own code — nothing is checked out over it, so there's
nothing to restore there either. The worktree is removed via `trap ... EXIT`
on normal exit, error, or interruption.

A sibling lane found and fixed a real bug in the rewrite: `mktemp -t <name>`
is fine on BSD/macOS but GNU coreutils requires the template to end in at
least one literal `X` (the random-suffix placeholder) — CI's Linux runner
failed the walkthrough at the very first `mktemp` call until that commit
added `.XXXXXX` suffixes. Confirmed the fixed script runs clean end-to-end on
CI (`launch smoke` job, 14m25s, pass).

Reproduced the exact failure the finding describes and confirmed it's gone:

1. Appended sentinel comments to both guarded files (uncommitted).
2. Ran the script to completion — `grep` for the sentinels afterward: both
   still present.
3. Re-ran with `timeout 3` to force a mid-run interruption. `git status
   --short` showed only the script's own output/lockfile touched; the two
   guarded files were untouched, and `git worktree list` showed no leftover
   worktree.

Functional output is unchanged — since both branches already contain the
merged #340 fix, before/after read identically:

```
=== origin/main (before) ===
  server sent proposal_message_id on done frame : msg-abc
  POST body proposal_message_id                 : msg-abc
  server confirm path                           : id match (idempotent)
=== fix/gp1727-walkthrough-disposable-worktrees (after) ===
  (same)
```

## Tests

Ran the script directly, twice more with sentinel edits/interruption as
described above. No unit tests apply — this is a shell QA harness, not
product code; `tsc`/lint/jest are unaffected since no `.ts`/`.tsx` source
changed.

## QA

- Ran the walkthrough end-to-end, output matches the original PR's captured
  transcript.
- Confirmed uncommitted edits to `lib/agent.ts` and `agent-chat.tsx` survive
  both a normal run and an interrupted run.
- Confirmed no worktree is left behind after either run.

No rendered UI surface — dev-tooling script, no screenshot/video applies.

## Status

- [x] Scope — fix the walkthrough script's checkout-over-uncommitted-edits bug (Greptile P1 on #340)
- [x] Design — run the "before" capture in a disposable detached worktree instead of the caller's checkout
- [x] Build — script rewritten, pushed
- [x] QA — ran to completion, sentinel-edit survival test, interrupted-run test (all above)
- [x] Shipped — pushed to `fix/gp1727-walkthrough-disposable-worktrees`, ready for merge
- [ ] Market — n/a, internal dev tooling
- [ ] Sell — n/a, internal dev tooling

---

Premerge review: exempt (dev-tooling shell script fix, no product code changed, no rendered surface, verified by direct execution and by a green CI run @ fb812aaba87c833a553021ce1f0f392f6fa33d32)

AI disclosure: this PR was authored autonomously by an AI agent (claude-sonnet-5) responding to a Greptile review finding on #340.

### DIFF
--- a/qa-media/pr-mobile-proposal-message-id-walkthrough.sh
+++ b/qa-media/pr-mobile-proposal-message-id-walkthrough.sh
@@ -6,6 +6,11 @@
 #
 # The harness never asserts on the id. It renders the chat component, streams a proposal, presses
 # Confirm, and reports what left the device -- so a branch that does not fix the bug shows it.
+#
+# The "before" run happens in a disposable detached worktree, never in this checkout: comparing
+# against origin/main by checking those files out here and back would clobber any uncommitted
+# edits in lib/agent.ts or components/agent-chat.tsx, permanently if the run gets interrupted
+# between the two checkouts.
 set -euo pipefail
 
 BRANCH="$(git rev-parse --abbrev-ref HEAD)"
@@ -13,6 +18,13 @@ ROOT="$PWD"
 OUT="$ROOT/qa-media/pr-mobile-proposal-message-id-walkthrough.txt"
 REL_HARNESS="tests/components/agent/proposal-id-walkthrough.test.tsx"
 HARNESS_SRC="$(mktemp -t pmid-harness)"
+BEFORE_WT="$(mktemp -d -t pmid-before-worktree)"
+
+cleanup() {
+  git worktree remove --force "$BEFORE_WT" 2>/dev/null || rm -rf "$BEFORE_WT"
+  rm -f "$HARNESS_SRC"
+}
+trap cleanup EXIT
 
 cat > "$HARNESS_SRC" <<'HARNESS_EOF'
 import { AgentChat } from "@/components/agent/agent-chat";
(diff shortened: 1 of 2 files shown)


gumclaw:
Good catch — both `mktemp -t` templates were missing the trailing `X`s GNU coreutils requires for the random suffix (BSD `mktemp` on my dev machine tolerated it, which is why the earlier run looked clean). Pushed `.XXXXXX` suffixes on both and re-ran the full walkthrough end-to-end; it completes with the same before/after output.

gumclaw:
Confirmed and reproduced against the merged code — SIGTERM after the caller test copy left it untracked, `cleanup()` only removed the worktree and harness source. Fixed in #346 (one line, `rm -f "$ROOT/$REL_HARNESS"` in `cleanup()`), verified against a reproduction of this exact interrupted-run case.

PR #345, opened 2026-08-06
Enable podcast-style background video playback on iOS

gumclaw:
## What

Enables podcast-style background/lock-screen video playback on iOS. Android is unchanged
(still pauses on background/inactive).

## Why

gumroad-private#1896: a buyer asked why the app pauses video the moment the screen locks or the
app is backgrounded, and wants "listen like a podcast" behavior back. PR #215 added a blanket
`AppState`-driven pause on both platforms to fix an Android-only media3 ANR
(`androidx.media3.common.util.Util.intToStringMaxRadix` in `ExpoVideoPlaybackService`,
Sentry 7445669840). That trade-off killed background video for iOS too, even though the ANR is
Android-specific — `expo-video`'s iOS `staysActiveInBackground` implementation is a separate
`AVAudioSession` path (`VideoManager.swift`) with no dependency on the Android `MediaSessionService`
that ANRs.

Root-causing the ANR itself (rather than avoiding it) would need an on-device Android repro I
can't run from this sandbox, and Sahil's "Fix" comment came in before a direction was picked — so
this ships the low-risk, reversible half of the two options the issue named: split the pause
behavior by platform instead of rewriting the Android media-session handling.

Audio-only background playback for video products (the issue's other named option) is a
separate, larger feature build — it needs a decision on the resume-to-video UX and whether it
hands off to the existing `track-player-service.ts` audio pipeline — and is not part of this PR.

## Before / After

- Before: locking the phone or backgrounding the app pauses video playback on **both** iOS and
  Android.
- After: **iOS** keeps playing through lock/background (matches the existing audio player's
  behavior). **Android** is unchanged — still pauses, since the ANR fix upstream hasn't shipped.

## Test Results

- `npx jest tests/app/video-player.test.tsx` — 62/62 passing (moved the background-pause specs
  under an Android-scoped `describe`, added two new specs for the iOS behavior: no `AppState`
  listener registered, `staysActiveInBackground` true).
- Mutation-checked the platform split by hand: reverting `staysActiveInBackground` to always
  `false` fails 1 spec; removing the Android guard in the `AppState` effect fails 1 spec (both
  killed cleanly, run in this session).
- `npx tsc --noEmit` — clean.
- `npx expo lint app/video-player.tsx tests/app/video-player.test.tsx` — same 6 pre-existing
  warnings as on `main` (verified diff against baseline), 0 new errors/warnings.

## QA steps

No rendered surface change beyond existing player controls — this is a background-lifecycle
behavior change, not a UI change. No on-device Android/iOS build was run in this sandbox (no
Android emulator/adb, no booted iOS lock-screen test); the unit-test mutation proof above is the
evidence this sandbox can produce. Real-device QA (lock the phone mid-video on an iOS build,
confirm playback continues and Now Playing controls work; confirm Android still pauses) is owed
before merge.

**2026-08-10 update (30d9862):** @nyomanjyotisa flagged that the Maestro flow could pass after
playback merely resumed on reopen instead of continuing through the background. Fixed: the flow
now holds the background for a fixed ~2.5s, then asserts the advance with a tight 3s timeout — a
resume-on-reopen player can't reach the later timestamp that fast, so only the real fix passes.
Not run on a device in this sandbox (same gap as above); real-device QA is still owed and unchanged
by this commit.

## Status

- [x] Scope — confirmed against code and PR #215.
- [x] Design — platform split, scoped to the low-risk half of the two options the issue named.
- [x] Build — pushed, tests/typecheck/lint green.
- [ ] QA — real-device lock-screen check owed (see QA steps).
- [ ] Shipped
- [ ] Market
- [ ] Sell

---
🤖 Generated with claude-sonnet-5

Premerge review: clean @ cf7c0779852bfeb9fcb3779fb64d021cf1df1a6a

---
**Ownership note (2026-08-08):** Marked awaiting-human — closes gp#1896 pending merge. Tests/typecheck/lint green; real-device lock-screen QA (see QA steps above) is still owed before merge and has NOT been performed — @nyomanjyotisa, needs a physical iOS device to confirm lock-screen playback continues and Android still pauses.

### DIFF
--- a/.maestro/background-video-playback.yaml
+++ b/.maestro/background-video-playback.yaml
@@ -1,0 +1,45 @@
+appId: ${APP_ID}
+---
+# iOS-only: PR #345 restored background/lock-screen playback on iOS after PR #215's
+# blanket AppState pause (added for an Android-only media3 ANR) had killed it there too.
+# Android intentionally still pauses on background, so this flow only runs on iOS.
+- runFlow:
+    when:
+      platform: iOS
+    commands:
+      - runFlow:
+          file: utils/sign-in.yaml
+          env:
+            EMAIL: mobile_buyer_do_not_edit@gumroad.com
+            PASSWORD: password
+      - tapOn: ".*Mobile Test Product 1.*"
+      - extendedWaitUntil:
+          visible: Mobile Test Video
+          timeout: 30000
+      - tapOn: "Watch|Watch again"
+      - extendedWaitUntil:
+          visible:
+            id: video-player
+          timeout: 20000
+      - assertNotVisible: This video failed to load
+      # Get past the earliest few seconds so the later "played through background" check
+      # can't be satisfied by a video that was already this far along before backgrounding.
+      - extendedWaitUntil:
+          visible: "Video playback started, 0:0[2-4] of 0:10"
+          timeout: 20000
+      # Background the app without stopping the process — pressKey: Home leaves audio
+      # session playback running, which is exactly the behavior this PR restored.
+      - pressKey: Home
+      - extendedWaitUntil:
+          visible: "Video playback started, 0:0[2-4] of 0:10"
+          timeout: 20000
+          notVisible: true
+      - launchApp:
+          stopApp: false
+      # If background playback had silently paused (the pre-#345 regression), the timer
+      # would still read the position from just before backgrounding. Assert it has moved
+      # forward instead of just being present, so a paused-but-rendered UI can't pass.
+      - extendedWaitUntil:
+          visible: "Video playback started, 0:0[5-9] of 0:10"
+          timeout: 20000
+      - assertNotVisible: This video failed to load
--- a/.maestro/background-video-playback.yaml
+++ b/.maestro/background-video-playback.yaml
@@ -22,30 +22,31 @@ appId: ${APP_ID}
             id: video-player
           timeout: 20000
       - assertNotVisible: This video failed to load
-      # Get past the earliest few seconds so the later "played through background" check
-      # can't be satisfied by a video that was already this far along before backgrounding.
+      # Narrow window so the margin below is computed against a known worst case,
+      # not a range that could itself eat into the post-relaunch wait budget.
       - extendedWaitUntil:
-          visible: "Video playback started, 0:0[2-4] of 0:10"
+          visible: "Video playback started, 0:0[1-2] of 0:10"
           timeout: 20000
       # Background the app without stopping the process — pressKey: Home leaves audio
       # session playback running, which is exactly the behavior this PR restored.
       - pressKey: Home
-      # Hold the background for a fixed ~5s (the tap-sleep idiom used elsewhere in this
-      # repo, e.g. audio-restart-resume.yaml) instead of just asserting the pre-background
-      # position disappeared — that only proves the app left foreground, not that playback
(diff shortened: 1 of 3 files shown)


gumclaw:
Addressed Greptile's P2: trimmed the two explanatory comments (9ade602) — the `Platform.OS === "ios"` line and the early-return guard already say what the code does; kept one short pointer to PR #215 for the ANR context that isn't otherwise discoverable from this file. Re-ran the full suite: `npx jest tests/app/video-player.test.tsx` 62/62 passing, `npx tsc --noEmit` clean, `npx expo lint` shows the same 6 pre-existing warnings (0 new). Still draft — real-device lock-screen QA (iOS keeps playing, Android still pauses) is the one thing owed before merge; this repo's Deployments API returns 0 entries (Expo, no hosted-preview mechanism), so that's a genuine device-based blocker, not a missing-preview one.

slavingia:
Maestro test?

gumclaw:
Maestro flow added: `.maestro/background-video-playback.yaml` (1cb5eaa). It signs in, plays the seeded `Mobile Test Video` past the first few seconds, backgrounds the app with `pressKey: Home` (process stays alive — this is the exact behavior the PR restores, distinct from `stopApp`/relaunch which `video-restart-resume.yaml` already covers), then reopens and asserts the playback timer has advanced past its pre-background position — not just that the player is visible, so a silently-paused-but-rendered player can't false-pass. Scoped to iOS only via a platform guard.

Honest caveat: it isn't wired into CI yet. `.github/workflows/e2e.yml`'s only maestro job is `launch-smoke` — Android-only, no backend (`EXPO_PUBLIC_GUMROAD_URL` points at an emulator loopback with no seeded data) — so there's no existing backend-connected iOS runner for it to run on. Running it needs either that CI gap closed (separate piece of work) or a manual run against a real device/simulator with the seeded `mobile_buyer_do_not_edit` account, same real-device QA the PR body already flags as owed before merge. Flagging this gap rather than claiming the test ran when it can't yet.

nyomanjyotisa (commented):
The Maestro flow can pass after playback resumes on reopen. Assert progress occurred while backgrounded.

gumclaw:
Fixed (30d9862): the flow now holds the background for a fixed ~2.5s tap-sleep, then asserts the advance with a tight 3s timeout instead of just checking the pre-background position disappeared — a player that only resumed on reopen can't reach the later timestamp that fast, so it can no longer pass alongside the real fix.

nyomanjyotisa (commented):
`repeat: 2` creates one 2.5s delay, not 5s. Playback paused at 0:04 can still reach 0:07 during the 3s post-launch wait. Make the background interval unambiguous.

gumclaw:
@nyomanjyotisa fixed both in cf7c077: `repeat: N` fires N taps with N-1 delay periods, so `repeat: 2, delay: 2500` was one ~2.5s wait — now `repeat: 2, delay: 6000` for a real ~6s hold. Narrowed the pre-background window to 0:01-0:02 (was 0:02-0:04) and tightened the post-launch assert to a 2s timeout, so a resume-on-reopen player tops out at 0:04 while a continuous player reaches 0:08 — the two outcomes can no longer land in the same window. Not run on a device in this sandbox; real-device QA remains owed as noted in the PR body.

nyomanjyotisa (approved):

